PR #15, opened 2021-06-21
Fix project initialization bugs

KonradSzwarc:
- Add `yarn` command to project setup instruction.
- Fix `theme` script in the `ui` package - move Inter.css import outside of the files that Chakra UI uses to generate theme.
- Unify `dev`, `build`, `start`, and `test` scripts across workspaces.
- Fix bug with running the server - change entry file path from the default (`dist/main`) to `dist/api/src/main`.
- Use ServeStaticModule only in the production environment.


packages/ui/package.json
```diff
@@ -3,9 +3,10 @@
   "version": "0.0.1",
   "private": true,
   "scripts": {
+    "dev": "start-storybook -p 6006",
```
hoolek77, line 6: Adjust script to start a storybook inside readme as well, because it's still `yarn w ui storybook`

README.md
```diff
@@ -11,9 +11,10 @@ Monorepo containing CodersCamp internal application and its website.
 ## How to set up the project?
 
 1. Clone the repository.
-2. Run `yarn bootstrap` command to install all dependencies and generate some necessary files.
-3. Run `docker-compose up -d` to start a database in the background.
-4. On Unix systems, run `chmod -R +x scripts ` to add execute permission to all scripts inside `scripts` directory.
+2. Run `yarn` to install all dependencies.
+3. On Unix systems, run `chmod -R +x scripts ` to add execute permission to all scripts inside `scripts` directory.
```
hoolek77, line 15: I don't think that space is necessary after bash command 😄

hoolek77, line 15: I don't think that space is necessary after the bash command 😄

kamildzierzak (approved):

hoolek77 (approved):

PR #16, opened 2021-06-23
feat: Create badge component

### LINKED ISSUE #11: Create Badge component
**Acceptance criteria:**
- Uses Badge component from Chakra UI under the hood.
- All its states match mockups.
- Its props extend `HTMLChakraProps<'div'>`.
- Receives the `size` prop with values `small`, `large`.
- Receives the `color` prop with values `default`, `brand`, `blue`, `green`, `orange`, `red`.
- Passes reference to the main element it renders.
- Has a playground story and stories that show its props combinations.

**Mockups:**
[Figma link](https://www.figma.com/file/ur0KLA9ZOtiodrAmpXrxh9/CodersCamp-app-and-website?node-id=10%3A983)

hoolek77:
- Main badge configuartion:
  - size: large or small
  - colors: each one matches mockups
  - variants: default, solid, or outlined (The default one matches mockups. Rest wasn't included in them, but I've decided to leave it anyway, just in case we need them in the future)
- PR also contains stories with playground and some basic tests

Closes #11

### DIFF
--- a/packages/ui/src/components/Badge/Badge.stories.tsx
+++ b/packages/ui/src/components/Badge/Badge.stories.tsx
@@ -17,7 +17,6 @@ export const Playground = Template.bind({});
 
 Playground.args = {
   children: 'Badge text',
-  variant: 'subtle',
   size: 'large',
   color: 'brand',
 };
@@ -41,17 +40,5 @@ export const Colors = () => (
     <Badge color="green">Green</Badge>
     <Badge color="orange">Orange</Badge>
     <Badge color="red">Red</Badge>
-  </HStack>
-);
-
-export const Variants = () => (
-  <HStack>
-    <Badge color="blue">Default</Badge>
-    <Badge variant="solid" color="blue">
-      Solid
-    </Badge>
-    <Badge variant="outline" color="blue">
-      Outline
-    </Badge>
   </HStack>
 );
--- a/packages/ui/src/components/Badge/Badge.stories.tsx
+++ b/packages/ui/src/components/Badge/Badge.stories.tsx
@@ -23,12 +23,8 @@ Playground.args = {
 
 export const Sizes = () => (
   <HStack>
-    <Badge size="small" color="red">
-      Small
-    </Badge>
-    <Badge size="large" color="red">
-      Large
-    </Badge>
+    <Badge size="small">Small</Badge>
+    <Badge size="large">Large</Badge>
   </HStack>
 );
 
--- a/packages/ui/src/components/Badge/Badge.stories.tsx
+++ b/packages/ui/src/components/Badge/Badge.stories.tsx
@@ -1,0 +1,57 @@
+import React from 'react';
+import { Meta, Story } from '@storybook/react';
+
+import { HStack } from '../Stack';
+import { Badge, BadgeProps } from './Badge';
+
+const meta: Meta = {
+  title: 'Badge',
+  component: Badge,
+};
+
+export default meta;
+
+const Template: Story<BadgeProps> = (args) => <Badge {...args} />;
+
+export const Playground = Template.bind({});
+
+Playground.args = {
+  children: 'Badge text',
+  variant: 'subtle',
+  size: 'large',
+  color: 'brand',
+};
+
+export const Sizes = () => (
+  <HStack>
+    <Badge size="small" color="red">
+      Small
+    </Badge>
+    <Badge size="large" color="red">
+      Large
+    </Badge>
+  </HStack>
+);
+
+export const Colors = () => (
+  <HStack>
+    <Badge color="default">Default</Badge>
+    <Badge color="brand">Brand</Badge>
+    <Badge color="blue">Blue</Badge>
+    <Badge color="green">Green</Badge>
+    <Badge color="orange">Orange</Badge>
+    <Badge color="red">Red</Badge>
+  </HStack>
+);
+
+export const Variants = () => (
+  <HStack>
+    <Badge color="blue">Default</Badge>
+    <Badge variant="solid" color="blue">
+      Solid
+    </Badge>
+    <Badge variant="outline" color="blue">
+      Outline
+    </Badge>
+  </HStack>
+);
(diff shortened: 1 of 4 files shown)


packages/ui/src/components/Badge/Badge.tsx
```diff
@@ -0,0 +1,57 @@
...
+type BadgeSize = 'small' | 'large';
+
+type BadgeColor = 'default' | 'brand' | 'blue' | 'green' | 'orange' | 'red';
+
+export interface BadgeProps extends HTMLChakraProps<'div'> {
+  children: ReactText | ReactText[];
+  /**
+   * Style variant of the badge
+   */
+  variant?: BadgeVariant;
```
KonradSzwarc, line 15: The badge shouldn't receive variant property. We don't know if we will use other variants of this component in the future so this property adds unnecessary complexity.

packages/ui/src/components/Badge/Badge.tsx
```diff
@@ -0,0 +1,57 @@
...
+  /**
+   * Color variant of the badge
+   */
+  color?: BadgeColor;
+  /**
+   * Determines badge's paddings, font size, line height and border radius.
+   */
+  size?: BadgeSize;
+  onClick?: ChakraBadgeProps['onClick'];
```
KonradSzwarc, line 24: For now, the badge shouldn't be clickable

packages/ui/src/components/Badge/Badge.tsx
```diff
@@ -0,0 +1,57 @@
...
+export const Badge = forwardRef<BadgeProps, 'div'>(
+  ({ variant = 'subtle', size = 'small', color = 'default', children, ...props }, ref) => {
+    return (
+      <ChakraBadge
+        variant={variant}
+        size={size}
+        colorScheme={color}
+        ref={ref}
+        fontWeight={500}
+        {...(size === 'small' ? { ...smallStyleProps } : { ...largeStyleProps })}
```
KonradSzwarc, line 50: It could more readable if you move the result of this conditional to a separate variable and only destructure it here

packages/ui/src/components/Badge/Badge.test.tsx
```diff
@@ -0,0 +1,35 @@
...
+  it('renders correctly in large size', () => {
+    const text = 'Badge Text';
+
+    render(
+      <Badge data-testid="badge" size="large">
+        {text}
+      </Badge>,
+    );
+
+    const badge = screen.getByTestId('badge');
```
KonradSzwarc, line 16: I think that `screen.getByText` will be a more suitable method to catch the Badge component. There is probably no need for the `data-testid` attribute here.

KonradSzwarc (changes requested):

KonradSzwarc:
I looked to the Storybook and badges mostly match mockups but there is one flaw. In Storybook Badge text is always written uppercase when in mockups it has a normal capitalization.
CleanShot 2021-06-28 at 18 15 46@2x

packages/ui/src/components/Badge/Badge.tsx
```diff
@@ -0,0 +1,50 @@
...
+};
+
+export const Badge = forwardRef<BadgeProps, 'div'>(({ size = 'small', color = 'default', children, ...props }, ref) => {
+  const sizeProps = size === 'small' ? { ...smallStyleProps } : { ...largeStyleProps };
+
+  return (
+    <ChakraBadge
+      textTransform="none"
+      size={size}
+      colorScheme={color}
```
KonradSzwarc, line 41: When `color` prop is equal to `'default'` you need to pass `gray` as the `colorScheme`. Currently, for the default badge, you have this CSS
CleanShot 2021-06-28 at 23 19 55@2x

packages/ui/src/components/Badge/Badge.tsx
```diff
@@ -0,0 +1,50 @@
...
+
+export const largeStyleProps = {
+  padding: '2px 12px',
+  borderRadius: '12px',
+  lineHeight: '20px',
+  fontSize: '14px',
+};
+
+export const Badge = forwardRef<BadgeProps, 'div'>(({ size = 'small', color = 'default', children, ...props }, ref) => {
+  const sizeProps = size === 'small' ? { ...smallStyleProps } : { ...largeStyleProps };
```
KonradSzwarc, line 35: I think there is no need to destructure `smallStyleProps` and `largeStyleProps` just to create an object from them again.

packages/ui/src/components/Badge/Badge.stories.tsx
```diff
@@ -0,0 +1,44 @@
...
+
+Playground.args = {
+  children: 'Badge text',
+  size: 'large',
+  color: 'brand',
+};
+
+export const Sizes = () => (
+  <HStack>
+    <Badge size="small" color="red">
```
KonradSzwarc, line 26: A small thing, but if we want to show how the component displays with different `size` prop, we shouldn't pass other props like `color`. I think you can just leave the default value here.

KonradSzwarc, line 26: A small thing, but if we want to show how the component displays with different `size` prop, we shouldn't pass other props like `color`. I think you can just leave the default value here.

KonradSzwarc (changes requested):

packages/ui/src/components/Badge/Badge.test.tsx
```diff
@@ -0,0 +1,27 @@
...
+describe('Badge', () => {
+  it('renders correctly in large size', () => {
+    const text = 'Badge Text';
+
+    render(<Badge size="large">{text}</Badge>);
+
+    const badge = screen.getByText(text);
+
+    expect(badge).toHaveStyle(largeStyleProps);
+    expect(badge).toBeVisible();
```
KonradSzwarc, line 15: When you want to check if an element correctly renders in the DOM it's recommended to use the `toBeInTheDocument` method

KonradSzwarc (approved):

KrystianKjjk (approved):

PR #17, opened 2021-06-24
Create workflow.yml

### LINKED ISSUE #6: Setup CI to run tests, types check and ESLint for each PR
**Acceptance criteria:**
For each PR there should be 3 workflows that always run:
- Lint - runs ESLint for all changed files when compared to the branch PR targets.
- Test - runs tests for all files affected by the changes when compared to the branch PR targets.
- Types check - runs TypeScript to check types for all files.
If any of those workflows fail (or will be in a pending state) the PR shouldn't be possible to merge.

KrystianKjjk:
Created workflow for: 
-tests, 
-es-lint, 
-typechecking

Closes #6


.github/workflows/workflow.yml
```diff
@@ -0,0 +1,65 @@
+# This workflow uses actions that are not certified by GitHub.
+# They are provided by a third-party and are governed by
+# separate terms of service, privacy policy, and support
+# documentation.
+
+# This workflow will install Deno and run tests across stable and canary builds on Windows, Ubuntu and macOS.
+# For more information see: https://github.com/denoland/setup-deno
```
KonradSzwarc, line 7: Are you sure we use Deno here? 😄

.github/workflows/workflow.yml
```diff
@@ -0,0 +1,65 @@
...
+      - name: install dependencies 
+        run: yarn install --immutable
+      - name: build api
+        run: yarn w api build
+      - name: build client
+        run: yarn w website build
+      - name: run affected tests 
+        run: yarn test --onlyChanged
+      
+  es-Lint:
```
KonradSzwarc, line 31: For consistency, it would be great to write all jobs keys in kebab-case

.github/workflows/workflow.yml
```diff
@@ -0,0 +1,65 @@
...
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v2
+      - name: install dependencies 
+        run: yarn install --immutable
+      - name: Run ESLint
+        run: yarn eslint --cache
+  Typescript-compiler:
```
KonradSzwarc, line 39: It might be nice to define testing, linting, and type checking as separate workflows so they can run concurrently

.github/workflows/workflow.yml
```diff
@@ -0,0 +1,65 @@
...
+    steps:
+    - uses: actions/checkout@v2
+    - name: install node v14
+      uses: actions/setup-node@v1
+      with:
+        node-version: 14
+    - name: install dependencies
+      run: yarn install
+    - name: tsc compile webside
+      uses: iCrawl/action-tsc@v1
```
KonradSzwarc, line 51: It looks like this action is no longer maintained (its repository is archived).
Maybe it would be better to define the `typecheck` command for each package and then add `yarn workspaces foreach -vpi run typecheck` script to the main `package.json` file. What do you think about it?

KonradSzwarc (changes requested):

KonradSzwarc:
I would suggest adding other reviewers to this PR, as you need at least two approves 😉

kamildzierzak (dismissed):

.github/workflows/workflow.yml
KrystianKjjk: done

kamildzierzak (approved):

hoolek77 (dismissed):

KonradSzwarc (approved):

KonradSzwarc (approved):

kamildzierzak (approved):

PR #18, opened 2021-06-27
feat(api): GitHub OAuth (#2)

### LINKED ISSUE #2: Implement "login with github" functionality
**Acceptance criteria:**
- User can click the sign-up button to be redirected to GitHub to log in.
- If the user doesn't exist we create a new user with the following properties:
```
id: number;
githubId: number;
firstName: string;
lastName: string;
email: string;
image: string;
```
- If the user exists we perform a login.
- After successful registration/login we redirect the user to the `/dashboard` page.

**Mockup (button)**
[Figma link](https://www.figma.com/file/ur0KLA9ZOtiodrAmpXrxh9/CodersCamp-app-and-website?node-id=3%3A219)

kacper-cyra:
User that sends request on path: "api/auth/github/login" is redirected to GitHub login page.
After successful login server receives user data.
If user does not exist he is being registered.
Token is created and returned together with user information.

Closes #2


.eslintrc.js
```diff
@@ -143,6 +142,11 @@ module.exports = {
       files: ['*.config.[tj]s', 'packages/website/pages/**/*.tsx', '*.stories.tsx'],
       rules: { 'import/no-default-export': 0 },
     },
+    {
+      // Enable eslint-plugin-testing-library rules only for test files
+      files: ['**/__tests__/**/*.[jt]s?(x)', '**/?(*.)+(spec|test).[jt]s?(x)'],
```
KonradSzwarc, line 147: That's a great idea to add this regex and use some plugins only for test files. I think we can make it a little better if we:
1. Define files only as `'**/?(*.)+(spec|test).ts?(x)'` as we don't have __tests__ directories and tests written as .js/.jsx files.
2. Move `'plugin:jest/recommended'` and `'plugin:jest-dom/recommended'` here as well.

packages/api/src/shared/shared.module.ts
```diff
@@ -0,0 +1,9 @@
+import { Module } from '@nestjs/common';
```
KonradSzwarc, line 1: I'm afraid that creating a module named Shared brings the risk it would be used for everything. I see you used it for configuration only so maybe  ConfigModule would be a better idea?

Also, I would consider making such a module global.

packages/api/src/auth/auth.service.ts
```diff
@@ -0,0 +1,23 @@
...
+import { UsersService } from '../users/users.service';
+import { GithubUserData } from './auth.model';
+
+@Injectable()
+export class AuthService {
+  constructor(private usersService: UsersService, private jwtService: JwtService) {}
+
+  login(user: User) {
+    const { id, ...profile } = user;
+    return this.jwtService.sign({ profile, sub: id });
```
KonradSzwarc, line 15: Why did you decide to store the entire user profile inside JWT? I consider if it wouldn't be enough if we store only the user's id.

packages/api/src/auth/strategies/github.client.ts
```diff
@@ -0,0 +1,25 @@
...
+import { ConfigService } from '../../shared/config.service';
+import { GithubDTO, GithubUserData } from '../auth.model';
+
+@Injectable()
+export class GithubClient extends PassportStrategy(Strategy) {
+  constructor(configService: ConfigService) {
+    super(configService.githubClientConfig);
+  }
+
+  async validate(_token: string, _refreshToken: string | undefined, profile: GithubDTO, done: VerifiedCallback) {
```
KonradSzwarc, line 15: It might be shorter to write `_refreshToken?: string` instead of `_refreshToken: string | undefined`

packages/api/src/users/users.service.ts
```diff
@@ -1,5 +1,8 @@
 import { Injectable } from '@nestjs/common';
 
+import { User } from '@coderscamp/shared/models/user';
+
+import { GithubUserData } from '../auth/auth.model';
```
KonradSzwarc, line 5: As those are only types you can write `import type` here. It's a great way to reduce the chance of circular dependencies error in the future.

packages/shared/src/models/user.ts
```diff
@@ -1,8 +1,10 @@
 export type User = {
   id: number;
-  firstName: string;
-  lastName: string;
+  githubId: number;
+  firstName: string | null;
```
KonradSzwarc, line 4: Maybe we can create `fullName` field instead of `firstName` + `lastName` here and use `displayName` provided by GitHub as an initial value?

KonradSzwarc (changes requested):

packages/api/src/shared/shared.module.ts
nowakprojects: I agree with Konrad. Sometimes Shared module is reasonable, but not in this case. 
We should prefer domain slices (vertical) instead of horizontal (technical) slices in our application architecture. 
So module per domain / feature. In this case better solution will be GithubClientConfigService / module etc. 

Then if you want to change something with Github changes will be introduced only in one module. 
Furthermore, let's imagine you want to remove possibility of GitHub login - then you remove a whole module and others remain the same. It's Open-Closed from SOLID in practice. Concerns which changes together should be coupled in some module.

packages/api/prisma/migrations/20210623204743_user/migration.sql
```diff
@@ -0,0 +1,3 @@
+-- AlterTable
```
nowakprojects, line 1: I'm not suer if migrations is right tool, if the app is not on prod. 
But I don't know best practices with prisma.

packages/api/src/auth/auth.model.ts
```diff
@@ -0,0 +1,16 @@
+export interface GithubDTO {
+  id: string; // numeric
+  nodeId: string;
+  displayName: string;
+  username: string;
+  profileUrl: string;
+  emails: { value: string }[];
```
nowakprojects, line 7: Why emails and photos are objects with property value inside whereas username / nodeId etc not?

packages/api/src/auth/auth.module.ts
```diff
@@ -0,0 +1,29 @@
...
+import { SharedModule } from '../shared/shared.module';
+import { UsersModule } from '../users/users.module';
+import { UsersService } from '../users/users.service';
+import { AuthController } from './auth.controller';
+import { AuthService } from './auth.service';
+import { GithubClient } from './strategies/github.client';
+
+@Module({
+  imports: [
+    PassportModule.register({ defaultStrategy: 'jwt' }),
```
nowakprojects, line 15: Is JWT strategy related to GitHub? Or it's totally different strategy?

packages/api/prisma/migrations/20210623204743_user/migration.sql
kacper-cyra: I will check it out

packages/api/src/auth/auth.model.ts
kacper-cyra: Well, that's the way passport.js maps response data. http://www.passportjs.org/docs/profile/

packages/api/src/auth/auth.module.ts
kacper-cyra: Yes, it's separate strategy

packages/api/src/auth/auth.model.ts
nowakprojects: Is it not strange, for you :) ? 

Ok, so here you have your auth domain model which is dependent on an external library. 
You should keep your domain model pure, because for example changing of library etc. should not change your logic etc. There is a place for simple mapping, IMO.

packages/api/src/auth/auth.module.ts
nowakprojects: So I don't see this separation in the modules. 
Let's ask yourself: Can JWT authorization exists without GitHub? And vice versa? 

Answer is YES ((maybe not, and I'm missing something?). So you should have two different modules - one for GitHub and one for internal / JWT strategy.

packages/api/src/auth/strategies/github.client.ts
nowakprojects: There is a difference between those two, as I understand it well.
With the question mark, you can ommit the parameter, with string | undefined you need to explicity tell that this param is undefined.

kacper-cyra:
I think that every comment has been addressed. I did make two separate modules inside of auth module: jwt and github.

pietruszka (dismissed):
Looks promising 👍 
Please see my comments

pietruszka:
@kacper-cyra FYI https://khalilstemmler.com/articles/typescript-domain-driven-design/repository-dto-mapper/

packages/api/.env
```diff
@@ -1 +1,8 @@
-DATABASE_URL="postgresql://root:root@localhost:5432/main-db"
\ No newline at end of file
+DATABASE_URL="postgresql://root:root@localhost:5432/main-db"
+
+GITHUB_CLIENT_ID = "e40b48326fd479c3dc42"
+GITHUB_CLIENT_SECRET = "5294c04782332725525316513f957cf657f858e7"
```
KonradSzwarc, line 4: It looks like something we don't want to have in our version control system. I would suggest leaving those two values as empty strings and writing in the docs, how to obtain them.

packages/api/package.json
```diff
@@ -11,7 +11,7 @@
     "test": "jest",
     "test:debug": "node --inspect-brk -r tsconfig-paths/register -r ts-node/register node_modules/.bin/jest --runInBand",
     "test:e2e": "jest --config ./test/jest-e2e.json",
-    "generate": "prisma migrate dev"
+    "generate": "prisma db push"
```
KonradSzwarc, line 14: What is the difference between the previous command and this one?

packages/api/src/auth/githubClient/github.client.ts
```diff
@@ -0,0 +1,43 @@
...
+import { PassportStrategy } from '@nestjs/passport';
+import { Strategy } from 'passport-github2';
+import { VerifiedCallback } from 'passport-jwt';
+
+import { UsersService } from '../../users/users.service';
+import { JwtStrategy } from '../jwtStrategy/jwt.strategy';
+import { GithubResponse, GithubUserData } from './github.model';
+
+@Injectable()
+export class GithubClient extends PassportStrategy(Strategy) {
```
KonradSzwarc, line 11: As this class extends `PassportStrategy`, I think `GithubStrategy` would be a more suitable name. This way we will avoid the strange path where you have `github.client` inside `githubClient`.
When it comes to naming I would rename `githubClient` directory to be simply `github`, but that's only a suggestion.

packages/api/src/auth/githubClient/github.client.ts
```diff
@@ -0,0 +1,43 @@
+import { Injectable } from '@nestjs/common';
+import { PassportStrategy } from '@nestjs/passport';
+import { Strategy } from 'passport-github2';
+import { VerifiedCallback } from 'passport-jwt';
+
+import { UsersService } from '../../users/users.service';
+import { JwtStrategy } from '../jwtStrategy/jwt.strategy';
+import { GithubResponse, GithubUserData } from './github.model';
```
KonradSzwarc, line 8: Add `type` keyword when you import only TS types.

packages/shared/src/models/user.ts
```diff
@@ -1,8 +1,10 @@
...
+  lastName: string | null;
   email: string;
+  image: string;
 };
 
 export type GetAllUsersResponse = User[];
```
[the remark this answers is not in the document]
KonradSzwarc: @pietruszka you're right. It was only created to demonstrate and test model sharing between server and client. It could be removed along with the entire related service, controller, and frontend handler. Saying so, I don't think it should be done in this PR as it would unnecessarily expand the scope of this task. I will create a separate issue for this.

KonradSzwarc (changes requested):

packages/api/package.json
kacper-cyra: Db push is used for schema prototyping and does not create history of changes, yet I think that we should still use prisma migrate each time we are done with modifying the schema.

kacper-cyra:
I made some changes and also I struggled to use postgres DB in memory. I found one solution: [pg-mem](https://github.com/oguimbal/pg-mem), but it has no support for Prisma yet. Prisma documentation suggests using docker for integration testing.

packages/api/.env.example
```diff
@@ -1,4 +1,11 @@
...
 DATABASE_URL="postgresql://root:root@localhost:5432/main-db"
-PORT=4000
\ No newline at end of file
+PORT=4000
+
+GITHUB_CLIENT_ID = ""
+GITHUB_CLIENT_SECRET = ""
+GITHUB_CALLBACK_URL = "http://localhost:4000/api/auth/github/callback"
+
+JWT_SECRET = ''
```
KonradSzwarc, line 10: It's a local secret key so I think we can preserve it here

packages/api/.env.example
```diff
@@ -1,4 +1,11 @@
...
+TOKEN_EXPIRATION_TIME = 3600
```
KonradSzwarc, line 11: Add a new line at the end of the file

packages/api/package.json
```diff
@@ -13,17 +13,23 @@
...
     "test:debug": "node --inspect-brk -r tsconfig-paths/register -r ts-node/register node_modules/.bin/jest --runInBand",
     "test:e2e": "jest --config ./test/jest-e2e.json",
-    "generate": "prisma generate"
+    "generate": "prisma db push"
   },
   "dependencies": {
     "@coderscamp/shared": "0.0.1",
     "@hapi/joi": "17.1.1",
     "@nestjs/common": "7.6.18",
+    "@nestjs/config": "0.6.3",
```
KonradSzwarc, line 22: There is probably no need for the config module as we have our `env.ts` file to validate environment variables and make them type-safe

packages/api/src/auth/github/github.client.ts
```diff
@@ -0,0 +1,42 @@
...
+import { UsersMapper } from '../../users/users.mapper';
+import { UsersRepository } from '../../users/users.repository';
+import { JwtStrategy } from '../jwtStrategy/jwt.strategy';
+import type { GithubResponse, GithubUserData } from './github.model';
+
+@Injectable()
+export class GithubClient extends GithubStrategy(Strategy) {
+  constructor(private usersRepository: UsersRepository, private jwtStrategy: JwtStrategy) {
+    super({
+      clientID: process.env.GITHUB_CLIENT_ID,
```
KonradSzwarc, line 16: Use environment variables from `env.ts` as they would be validated and type-safe.

packages/api/src/auth/github/github.model.ts
```diff
@@ -0,0 +1,18 @@
+/* eslint-disable @typescript-eslint/naming-convention */
```
KonradSzwarc, line 1: What is the problem with our naming convention. I think we should rather update our ESLint config than disable this rule.

packages/api/src/auth/jwtStrategy/jwt.module.ts
```diff
@@ -0,0 +1,21 @@
+import { Module } from '@nestjs/common';
+import { ConfigModule, ConfigService } from '@nestjs/config';
+import { JwtModule } from '@nestjs/jwt';
+
+import { JwtStrategy } from './jwt.strategy';
+
+@Module({
+  imports: [
+    JwtModule.registerAsync({
```
KonradSzwarc, line 9: Use values from the `env.ts`. This way it would be more type-safe and you wouldn't need to register this module asynchronously.

packages/api/src/users/users.mapper.ts
```diff
@@ -0,0 +1,32 @@
+import type { User } from '@coderscamp/shared/models/user';
+
+import type { GithubDTO } from '../auth/github/github.model';
+import type { UserDTO } from './users.model';
+
+export class UsersMapper {
+  static toDomain(value: UserDTO): User {
+    return value as User;
```
KonradSzwarc, line 8: Why do we need to cast type in this and nearly all other methods? It doesn't seem right as we might lose type-safety.

packages/api/src/users/users.controller.ts
```diff
@@ -1,15 +1,15 @@
...
 
 @Controller('users')
 export class UsersController {
-  constructor(private readonly usersService: UsersService) {}
+  constructor(private readonly usersRepository: UsersRepository) {}
 
   @Get('/')
-  async getAll(): Promise<GetAllUsersResponse> {
-    return this.usersService.getAll();
+  async getAll(): Promise<UserDTO[]> {
```
KonradSzwarc, line 12: The return type of the controller should be placed in the shared package (like GetAllUsersResponse was before) so we can share it with the frontend.

packages/api/src/users/users.model.ts
```diff
@@ -0,0 +1,7 @@
+export interface UserDTO {
```
KonradSzwarc, line 1: We have a `shared/src/models` directory to define models. This way they could be shared between many apps

packages/api/src/users/users.repository.ts
```diff
@@ -0,0 +1,35 @@
+import { Injectable } from '@nestjs/common';
+
+import type { User } from '@coderscamp/shared/models/user';
+
+import { PrismaService } from '../prisma/prisma.service';
+
+@Injectable()
+export class UsersRepository {
```
KonradSzwarc, line 8: Shouldn't it be rather a service than a repository?

packages/api/src/users/users.repository.ts
```diff
@@ -0,0 +1,35 @@
...
+      where: { id },
+      data,
+    });
+  }
+
+  async getById(id: number): Promise<User | null> {
+    return this.prisma.user.findUnique({
+      where: { id },
+    });
+  }
```
KonradSzwarc, line 30: Are those methods used anywhere in the app?

packages/api/src/auth/github/github.client.ts
```diff
@@ -0,0 +1,42 @@
...
+    _refreshToken: string | undefined,
+    { _json: profile }: GithubResponse,
+    done: VerifiedCallback,
+  ) {
+    done(null, UsersMapper.fromGithubInputToDomain(profile));
+  }
+
+  async githubAuth(user: GithubUserData): Promise<{ accessToken: string; profile: UserDTO } | null> {
+    let userFromDatabase = await this.usersRepository.getByGithubId(user.githubId);
+    if (!userFromDatabase) userFromDatabase = await this.usersRepository.create(user);
```
KonradSzwarc, line 34: Use brackets when you define an `if` statement. This way it would be easier to read

KonradSzwarc (changes requested):

packages/api/src/auth/github/github.model.ts
kacper-cyra: Eslint at this moment does not accept property name that comes from github: avatar_url

KonradSzwarc: In such a case add `// eslint-disable-next-line @typescript-eslint/naming-convention` to disable only the line with wrongly named property.

packages/api/src/users/users.repository.ts
nowakprojects: Looks like repository for me. Just data retrieving / updating.

packages/api/src/users/users.mapper.ts
pietruszka: This mapper function doesn't have any sense. You are casting and you do not validate/normalize any value.
Please read, https://docs.nestjs.com/techniques/serialization
https://docs.nestjs.com/pipes#class-validator 
You can also use `class-transformer-validator`. 
As example 
```
  static toDomain(
    props: Partial<UserCore>,
    options: { keyCode?: string } = {},
  ): User {
    const createUser = options.keyCode ? User.create : User.recreate;

    return createUser(
      transformAndValidateSync(UserCore, props, {
        transformer: { excludeExtraneousValues: true },
        validator: { validationError: { target: false, value: false } },
      }),
      options.keyCode,
    );
  }

  static toPlain(child: User): UserCore {
    const props = child.getProps();

    return classToPlain(props, { excludeExtraneousValues: true }) as UserCore;
  }```

File: .eslintrc.js [hunk 27]
```diff
@@ -67,7 +67,14 @@ module.exports = {
     // Enforces some naming conventions across the codebase.
     '@typescript-eslint/naming-convention': [
       2,
-      { selector: 'default', format: ['camelCase', 'UPPER_CASE', 'PascalCase'] },
+      {
+        selector: 'default',
+        format: ['camelCase', 'UPPER_CASE', 'PascalCase'],
+        filter: {
+          regex: '^(avatar_url|_json)$',
```
KonradSzwarc, line 74: I think using `eslint-disable-next-line @typescript-eslint/naming-convention` in places you have those variables would be a better solution. If we one day decide to remove one of those variables no one will remember that there is a related ESLint rule we need to modify as well.

KonradSzwarc (commented):
When it comes to the main feature, everything looks great for me ✅
There is only one thing to update is ESLint.

.eslintrc.js
kacper-cyra: Okay, so I did as suggested.

README.md
```diff
@@ -2,5 +2,140 @@
 
 Monorepo containing CodersCamp internal application and its website.
 
+## Project requirements
```
KonradSzwarc, line 5: 🤔 You should probably merge the main branch, as this documentation shouldn't be here anymore

KonradSzwarc (changes requested):
There is only a problem with README.md which content should be already moved to the docs. When it comes to the functionality in my opinion everything is great 👍

README.md
kacper-cyra: Done

KonradSzwarc (dismissed):

KonradSzwarc (dismissed):

hoolek77 (approved):

KonradSzwarc (dismissed):

KonradSzwarc (approved):

hoolek77 (approved):

PR #24, opened 2021-06-29
feat: Feature 20 typography component (#20)

### LINKED ISSUE #20: Create Typography component
**Acceptance criteria:**
- Uses Box component from Chakra UI under the hood.
- All its states match mockups.
- Its props extend HTMLChakraProps<'div'>.
- Receives the `size` prop with values from `xs` to `7xl`.
- Receives the `weight` prop with values `regular` and `medium`.
- Receives the `as` prop with values `span`, `div`, `a`, `h1`-`h6`, `p`.
- Passes reference to the main element it renders.
- Has a playground story and stories that show its props combinations.
- When `as` prop is equal to `a` the component should receive the `href` prop. In such a case, the rendered text should have `brand.500` (`brand.600` when hovered and `brand.700` when active) color and be underlined.

lagdush:
Receives the size prop with values from **xs** to **7xl**.
Receives the weight prop with values **regular** and **medium**.
Receives the as prop with values **span, div, a, h1-h6, p.**
Component have stories and basic test.
Closes #20


packages/ui/src/components/Typography/Typography.stories.tsx
```diff
@@ -0,0 +1,90 @@
+import React from 'react';
+import { Meta, Story } from '@storybook/react';
+
+import { Typography, BoxProps } from './Typography';
```
KonradSzwarc, line 4: Add a new line after imports. It should be fixed automatically by ESLint, so you might have some issue with its configuration across your IDE

packages/ui/src/components/Typography/Typography.stories.tsx
```diff
@@ -0,0 +1,90 @@
...
+    </Typography>
+  </>
+);
+
+export const Variants = () => (
+  <>
+    <Typography size="md" as="p">
+      Paragraph
+    </Typography>
+    <Typography size="md" as="a" href="#">
```
KonradSzwarc, line 66: It's my fault because I didn't specify that in the issue, but for anchors text should be underlined and has the brand color. I wrote it down, so now you can find this requirement as the last point in the issue.

packages/ui/src/components/Typography/Typography.stories.tsx
```diff
@@ -0,0 +1,90 @@
...
+      Span
+    </Typography>
+  </>
+);
+
+export const FontWeights = () => (
+  <>
+    <Typography size="lg" as="p" weight='medium'>
+      Medium Weight
+    </Typography>{' '}
```
KonradSzwarc, line 85: Why we need the `{' '}` at the end of this line?

packages/ui/src/components/Typography/Typography.tsx
```diff
@@ -0,0 +1,25 @@
+import React, { ReactNode, ReactText } from 'react';
+import { Box, forwardRef, HTMLChakraProps } from '@chakra-ui/react';
+
+type TypographyVariant = 'span' | 'div' | 'a' | 'p' | 'h1' | 'h2' | 'h3' | 'h4' | 'h5' | 'h6';
+
+type TypographyFontSize = 'xs' | 'sm' | 'md' | 'lg' | 'xl' | '2xl' | '3xl' | '4xl' | '5xl' | '6xl' | '7xl';
+
+type TypographyFontWeight = 'regular' | 'medium';
+
+export interface BoxProps extends HTMLChakraProps<'div'> {
```
KonradSzwarc, line 10: It's a Typography component so it would be better to name this interface `TypographyProps`

packages/ui/src/components/Typography/index.ts
```diff
@@ -0,0 +1,2 @@
+export { Typography } from './Typography';
+export type { TypographyProps } from './Typography';
```
KonradSzwarc, line 2: This interface doesn't exist

KonradSzwarc (changes requested):

packages/ui/src/components/Typography/Typography.stories.tsx
lagdush: Yes, I had a problem with the IDE and ESLint but I think I solved it;-)

packages/ui/src/components/Typography/Typography.tsx
```diff
@@ -0,0 +1,48 @@
...
+type TypographyFontWeight = 'regular' | 'medium';
+
+type RemoveCommonValues<Type, TOmit> = {
+  [Property in keyof Type]: TOmit extends Record<Property, infer U> ? Exclude<Type[Property], U> : Type[Property];
+};
+
+type Id<Type> = Record<string, unknown> & { [P in keyof Type]: Type[P] };
+type ConditionalProps<Type, TKey extends keyof TCase, TCase extends Partial<Type>> =
+  | Id<Omit<Type, keyof TCase> & TCase>
+  | Id<RemoveCommonValues<Type, Pick<TCase, TKey>>>;
```
KrystianKjjk, line 17: I would recommend  to put RemoveCommonValues and ConditionalProps in other file (in shared forlder maybe?), because if anybody else will need that, it would be easy to use it then

packages/ui/src/components/Typography/Typography.stories.tsx
```diff
@@ -0,0 +1,96 @@
+import React from 'react';
+import { Meta, Story } from '@storybook/react';
+
+import { HStack } from '../Stack';
+import { IsAnchor, Typography } from './index';
+
+const tags = { h1: 'h1', a: 'a' };
```
KonradSzwarc, line 7: I think we could specify all possible tags here

packages/ui/src/components/Typography/Typography.stories.tsx
```diff
@@ -0,0 +1,96 @@
...
+Playground.args = {
+  children: 'lorem ipsum dolor sit amet',
+  as: 'h1',
+  size: '2xl',
+  weight: 'regular',
+};
+
+export const Sizes = () => (
+  <>
+    <Typography size="xs" as="h1">
```
KonradSzwarc, line 37: If you want to demonstrate the `size` prop you shouldn't use any other props

packages/ui/src/components/Typography/Typography.tsx
```diff
@@ -0,0 +1,48 @@
...
+  as?: TypographyVariant | 'a';
+  size?: TypographyFontSize;
+  weight?: TypographyFontWeight;
+}
+export interface TypographyPropsForAnchor extends TypographyProps {
+  href: string;
+  as: 'a';
+}
+
+export type IsAnchor = ConditionalProps<TypographyProps, 'as', TypographyPropsForAnchor>;
```
KonradSzwarc, line 30: I think you should save this type as a gist, as it might be useful for some components.
Saying so, there is no need to use it here as the previous solution worked really well. We don't want to require the `href` property as sometimes it would be provided by the parent component (e.g. [Next Link](https://nextjs.org/docs/api-reference/next/link)). I checked and even for the previous solution, I have IntelliSense for `href` and `target` props when using `as="a"` prop.

packages/ui/src/components/Typography/Typography.tsx
```diff
@@ -0,0 +1,48 @@
...
+  [Property in keyof Type]: TOmit extends Record<Property, infer U> ? Exclude<Type[Property], U> : Type[Property];
+};
+
+type Id<Type> = Record<string, unknown> & { [P in keyof Type]: Type[P] };
+type ConditionalProps<Type, TKey extends keyof TCase, TCase extends Partial<Type>> =
+  | Id<Omit<Type, keyof TCase> & TCase>
+  | Id<RemoveCommonValues<Type, Pick<TCase, TKey>>>;
+
+export interface TypographyProps extends HTMLChakraProps<'div'> {
+  children: ReactText | ReactText[] | ReactNode | ReactNode[];
```
KonradSzwarc, line 20: You can simply leave `ReactNode` here as it includes all other types.

packages/ui/src/components/Typography/Typography.tsx
```diff
@@ -0,0 +1,48 @@
...
+  textDecoration: 'underline',
+  _hover: { color: 'brand.600' },
+  _active: { color: 'brand.700' },
+};
+
+export const Typography: React.FC<IsAnchor> = forwardRef(
+  ({ as = 'div', size = 'md', weight = 'regular', children, ...props }, ref) => {
+    const stylesForLink = as === 'a' && anchorStyle;
+    return (
+      <Box as={as} fontSize={size} ref={ref} fontWeight={weight} {...props} {...stylesForLink}>
```
KonradSzwarc, line 43: Wait, `stylesForLink` could be an object or `false` so in many cases you're destructuring the `false` value. I see it works without any error, but I'm thinking why? 😅

packages/ui/src/components/Typography/Typography.stories.tsx
```diff
@@ -0,0 +1,96 @@
+import React from 'react';
+import { Meta, Story } from '@storybook/react';
+
+import { HStack } from '../Stack';
+import { IsAnchor, Typography } from './index';
```
KonradSzwarc, line 5: Import from the `./Typografy.tsx` file as it's an easier way to protect the app from dependency cycles.

KonradSzwarc (changes requested):

packages/ui/src/components/Typography/Typography.tsx
lagdush: The magic of JavaScript;-) I can change it to a ternary operator.

packages/ui/src/components/Typography/Typography.tsx
lagdush: Should I change the component type to the previous one here,   
  ```
children: ReactNode;
  as?: TypographyVariant | 'a';
  size?: TypographyFontSize;
  weight?: TypographyFontWeight;
``` and add optional href as string?

Comment (KonradSzwarc) [reply packages/ui/src/components/Typography/Typography.tsx:30]: Something like this
```
type TypographyVariant = 'span' | 'div' | 'a' | 'p' | 'h1' | 'h2' | 'h3' | 'h4' | 'h5' | 'h6';

export interface TypographyProps extends HTMLChakraProps<'div'> {
  children: ReactNode;
  as?: TypographyVariant;
  size?: TypographyFontSize;
  weight?: TypographyFontWeight;
}
```

There is no need to specify the `href` property. As you extend `HTMLChakraProps` this component would have `href` in autosuggestions if someone provide `as="a"` in props.

File: packages/ui/src/components/Typography/Typography.tsx (hunk 11 above)
Comment (KonradSzwarc) [reply packages/ui/src/components/Typography/Typography.tsx:43]: We can discuss it on Mondays daily. For me, your current solution works fine and it's shorter so you can leave it 😄

Comment (KonradSzwarc) [reply packages/ui/src/components/Typography/Typography.tsx:43]: As a curiosity, there is the reason, why it works. It's a result of how Babel transpiles spread operators
CleanShot 2021-07-03 at 22 05 53@2x

File: packages/ui/src/components/Typography/Typography.stories.tsx [hunk 13]
```diff
@@ -0,0 +1,74 @@
...
+    <Typography size="4xl">Extra Large (4xl)</Typography>
+    <Typography size="5xl">Extra Large (5xl)</Typography>
+    <Typography size="6xl">Extra Large (6xl)</Typography>
+    <Typography size="7xl">Extra Large (7xl)</Typography>
+  </>
+);
+
+export const Variants = () => (
+  <HStack>
+    <Typography size="md" as="span">
```
KonradSzwarc, line 53: Why those variants have different sizes?

KonradSzwarc (approved):

KrystianKjjk (dismissed):

packages/ui/src/components/Typography/Typography.stories.tsx
lagdush: My mistake I will correct it;-)

packages/ui/src/components/Typography/Typography.stories.tsx
```diff
@@ -0,0 +1,74 @@
...
+      Span
+    </Typography>
+    <Typography size="md" as="a" href="#">
+      Link
+    </Typography>
+    <Typography size="md" as="p">
```
KonradSzwarc, line 59: I think there is no need to pass the `size` prop when you use it as the default value

packages/ui/src/components/Typography/Typography.stories.tsx
```diff
@@ -0,0 +1,74 @@
...
+      Paragraph
+    </Typography>
+  </HStack>
+);
+
+export const FontWeights = () => (
+  <>
+    <Typography size="lg" as="p" weight="medium">
```
KonradSzwarc, line 67: If you want to compare weights pass only the `weight` prop

KonradSzwarc (approved):

KonradSzwarc (approved):

hoolek77 (approved):

PR #43, opened 2021-07-07
feat(ui): Create NumberInput Component (#32)

### LINKED ISSUE #32: Create NumberInput component
**Acceptance criteria:**
- Uses NumberInput component from Chakra UI under the hood.
- It's props extends `Omit<HTMLChakraProps<"div">, keyof UseNumberInputProps>`.
- Receives the following props inherited directly from Chakra UI: all aria-* attributes, value, onChange, name, max, min, precision, step.
- Restricts the original Chakra's input `size` prop values to `sm`, `md` and `lg`.
- Restricts the original Chakra's input `inputMode ` prop values to `numeric` and `decimal`.

**Mockups:**
[Number Input](https://www.figma.com/file/ur0KLA9ZOtiodrAmpXrxh9/CodersCamp-app-and-website?node-id=43%3A11)

talcia:
Closes #32


packages/ui/src/components/NumberInput/NumberInput.tsx
```diff
@@ -0,0 +1,118 @@
...
+} from '@chakra-ui/react';
+
+type NumberInputSize = 'sm' | 'md' | 'lg';
+
+type NumberInputMode = 'numeric' | 'decimal';
+
+export interface NumberInputProps extends Omit<HTMLChakraProps<'div'>, keyof UseNumberInputProps> {
+  size: NumberInputSize;
+  inputMode: NumberInputMode;
+  disabled: boolean;
```
hoolek77, line 20: I think it would be easier to use this component if input mode and disabled props were not required. And while destructuring them, set the default value for input mode to decimal and for the disabled prop to false.

packages/ui/src/components/NumberInput/NumberInput.tsx
```diff
@@ -0,0 +1,118 @@
...
+      'aria-describedby': ariaDescribedby,
+      'aria-label': ariaLabel,
+      'aria-labelledby': ariaLabelledby,
+      value,
+      name,
+      max,
+      min,
+      precision,
+      onChange,
+      step,
```
hoolek77, line 59: You don't have to destructure these props, just spread the rest of them like: 
`...props`
and then pass it to the chakra component

hoolek77 (changes requested):

packages/ui/src/components/NumberInput/NumberInput.tsx
hoolek77: Also, add short descriptions for the first three props. Check other components how it should be done.

packages/ui/src/components/NumberInput/NumberInput.tsx
talcia: I tried do this like you recommend but I got an error on ChakraNumberInput about invalid type of props. So I decided to destructure this props

packages/ui/src/components/NumberInput/NumberInput.stories.tsx
```diff
@@ -0,0 +1,52 @@
...
+
+Playground.args = {
+  size: 'sm',
+  inputMode: 'decimal',
+  disabled: false,
+};
+
+export const Sizes = () => (
+  <HStack>
+    <NumberInput size="sm" inputMode="decimal" disabled={false} />
```
KonradSzwarc, line 26: For stories showing different sizes of the NumberInput, you should focus only on the `size` prop. `inputMode` and `disabled` are no needed. This applies to other stories as well.

packages/ui/src/components/NumberInput/NumberInput.stories.tsx
```diff
@@ -0,0 +1,52 @@
...
+  <HStack>
+    <NumberInput size="md" inputMode="decimal" disabled={false} />
+    <NumberInput size="md" inputMode="numeric" disabled={false} />
+  </HStack>
+);
+
+export const Disabled = () => (
+  <HStack>
+    <NumberInput size="md" inputMode="decimal" disabled />
+    <NumberInput size="md" inputMode="numeric" disabled={false} />
```
KonradSzwarc, line 42: `disabled={false}` doesn't seem right. As `false` is the default value it should be omitted when we don't want the component to be disabled.

packages/ui/src/components/NumberInput/NumberInput.tsx
```diff
@@ -0,0 +1,127 @@
...
+
+export const mdStyleProps = {
+  height: '40px',
+  padding: '1px 0px 1px 16px',
+};
+
+export const smStyleProps = {
+  height: '32px',
+  padding: '1px 0px 1px 12px',
+};
```
KonradSzwarc, line 51: This should be specified on a theme level. You can find more [here](https://chakra-ui.com/docs/theming/component-style).

packages/ui/src/components/NumberInput/NumberInput.tsx
```diff
@@ -0,0 +1,127 @@
...
+  /**
+   * Determines if number input is disabled or not .
+   */
+  disabled?: boolean;
+  value?: number;
+  onChange?: (arg: string) => void;
+  max?: number;
+  min?: number;
+  precision?: number;
+  step?: number;
```
KonradSzwarc, line 35: It would be great to add descriptions for those props as well

packages/ui/src/components/NumberInput/NumberInput.tsx
```diff
@@ -0,0 +1,127 @@
...
+  /**
+   * Determines type of data that might be entered by the user (numeric or decimal).
+   */
+  inputMode?: NumberInputMode;
+  /**
+   * Determines if number input is disabled or not .
+   */
+  disabled?: boolean;
+  value?: number;
+  onChange?: (arg: string) => void;
```
KonradSzwarc, line 31: Correct type for this function is `(stringValue: string, numberValue: number) => void`

packages/ui/src/components/NumberInput/NumberInput.tsx
```diff
@@ -0,0 +1,127 @@
...
+      border: '1px solid #9CA3AF',
+    };
+
+    const focusStyleProps = {
+      'box-shadow': '0px 0px 0px 3px rgba(56, 189, 248, 0.6)',
+    };
+
+    const invalidStyleProps = {
+      border: '2px solid #EF4444',
+    };
```
KonradSzwarc, line 93: All those styles probably could be moved to the theme styling of the NumberInput

packages/ui/src/components/NumberInput/NumberInput.tsx
KonradSzwarc: If you can, move back to the version with spread and paste an error you're getting. Destructuring so many props makes this component a little harder to read. Of course, if there is no other way we will leave it, but it's still worth trying to make it cleaner.

packages/ui/src/components/NumberInput/NumberInput.test.tsx
```diff
@@ -0,0 +1,49 @@
+import React from 'react';
+import { render, screen } from '@testing-library/react';
+
+import { lgStyleProps, mdStyleProps, NumberInput, smStyleProps } from './NumberInput';
+
+describe('NumberInput', () => {
+  it('renders correctly in not disabled mode', () => {
+    render(<NumberInput size="lg" inputMode="numeric" disabled={false} aria-label="inputNumber" />);
```
KonradSzwarc, line 8: Don't add props that wouldn't affect the test to the component. In this case, only the `aria-label` prop is needed for this test to work properly

KonradSzwarc (changes requested):

packages/ui/src/components/NumberInput/NumberInput.stories.tsx
```diff
@@ -0,0 +1,52 @@
...
+  <HStack>
+    <NumberInput size="sm" />
+    <NumberInput size="md" />
+    <NumberInput size="lg" />
+  </HStack>
+);
+
+export const InputModes = () => (
+  <HStack>
+    <NumberInput size="md" inputMode="decimal" />
```
KonradSzwarc, line 34: As `md` is the default size there is no need to provide it explicitly

packages/ui/src/components/NumberInput/NumberInput.stories.tsx
```diff
@@ -0,0 +1,52 @@
...
+export const Disabled = () => (
+  <HStack>
+    <NumberInput size="md" disabled />
+    <NumberInput size="md" />
+  </HStack>
+);
+
+export const Pseudoclasses = () => (
+  <HStack>
+    <NumberInput size="md" inputMode="decimal" disabled={false} />
```
KonradSzwarc, line 48: `size`, `inputMode`, and `disabled` props probably shouldn't be here

packages/ui/src/components/NumberInput/NumberInput.test.tsx
```diff
@@ -0,0 +1,49 @@
+import React from 'react';
+import { render, screen } from '@testing-library/react';
+
+import { lgStyleProps, mdStyleProps, NumberInput, smStyleProps } from './NumberInput';
+
+describe('NumberInput', () => {
+  it('renders correctly in not disabled mode', () => {
+    render(<NumberInput size="lg" aria-label="inputNumber" />);
```
KonradSzwarc, line 8: The `size` prop doesn't affect this test so there is no need to provide it

packages/ui/src/components/NumberInput/NumberInput.test.tsx
```diff
@@ -0,0 +1,49 @@
...
+
+    const numberInput = screen.getByLabelText('inputNumber');
+
+    expect(numberInput).toBeEnabled();
+    expect(numberInput).toBeInTheDocument();
+  });
+
+  it('renders correctly in disabled mode', () => {
+    render(<NumberInput size="lg" disabled aria-label="inputNumber" />);
```
KonradSzwarc, line 17: The `size` prop doesn't affect this test so there is no need to provide it

packages/ui/src/components/NumberInput/NumberInput.test.tsx
```diff
@@ -0,0 +1,49 @@
+import React from 'react';
+import { render, screen } from '@testing-library/react';
+
+import { lgStyleProps, mdStyleProps, NumberInput, smStyleProps } from './NumberInput';
+
+describe('NumberInput', () => {
+  it('renders correctly in not disabled mode', () => {
+    render(<NumberInput size="lg" aria-label="inputNumber" />);
+
+    const numberInput = screen.getByLabelText('inputNumber');
```
Veryyapeee, line 10: (Minor)
You could make function for this instead of calling screen method every time
const getElement = () => screen.getByLabelText('inputNumber')

KonradSzwarc: It can be done but only if we move the entire `render` method to that function or if the function takes the `ariaLabel` parameter. Otherwise, the function you suggest will depend on the aria-label value that is outside of the scope of that function.
As you write it's a minor thing so in my opinion it could be addressed but doesn't have to.

packages/ui/src/components/NumberInput/NumberInput.tsx
```diff
@@ -0,0 +1,70 @@
...
+      <ChakraNumberInput
+        size={size}
+        inputMode={inputMode}
+        isDisabled={disabled}
+        aria-describedby={ariaDescribedby}
+        aria-label={ariaLabel}
+        aria-labelledby={ariaLabelledby}
+        ref={ref}
+        {...props}
+      >
```
talcia, line 61: Type '{ children: Element[]; form?: string | undefined; pattern?: string | undefined; defaultValue?: string | number | readonly string[] | undefined; id?: string | undefined; onFocus?: FocusEventHandler<HTMLInputElement> | undefined; ... 659 more ...; ref: ForwardedRef<...>; }' is not assignable to type 'NumberInputProps'.
  Types of property 'defaultValue' are incompatible.
    Type 'string | number | readonly string[] | undefined' is not assignable to type 'StringOrNumber | undefined'.
      Type 'readonly string[]' is not assignable to type 'StringOrNumber | undefined'.
        Type 'readonly string[]' is not assignable to type 'string'.ts(2322)
(alias) const ChakraNumberInput: ComponentWithAs<"div", NumberInputProps>
import ChakraNumberInput
NumberInput

React component that provides context and logic to all number input sub-components.

image

packages/ui/src/theme/overwrites/components/numberInput.ts
```diff
@@ -0,0 +1,38 @@
...
+  },
+  variants: {
+    outline: {
+      field: {
+        borderRadius: '6px',
+        _focus: {
+          boxShadow: '0px 0px 0px 3px rgba(56, 189, 248, 0.6)',
+        },
+        _hover: {
+          border: '1px solid #9CA3AF',
```
KonradSzwarc, line 30: 🤔 Is there a need to define border type and width? Maybe just specifying the `borderColor` property would be enough? Also, it would be great if you use the name of the color from our theme instead of a hex here

packages/ui/src/theme/overwrites/components/numberInput.ts
```diff
@@ -0,0 +1,38 @@
...
+        padding: '1px 0px 1px 16px',
+      },
+    },
+  },
+  variants: {
+    outline: {
+      field: {
+        borderRadius: '6px',
+        _focus: {
+          boxShadow: '0px 0px 0px 3px rgba(56, 189, 248, 0.6)',
```
KonradSzwarc, line 27: 🚨 This shadow is already defined in our theme. Its name is 'outlined'

packages/ui/src/theme/overwrites/components/numberInput.ts
```diff
@@ -0,0 +1,38 @@
+export const NumberInput = {
+  sizes: {
```
KonradSzwarc, line 2: 🤔 Doesn't NumberInput sizes match mockups by default? In other words, is there a need to customize them?

packages/ui/src/theme/overwrites/components/numberInput.ts
```diff
@@ -0,0 +1,38 @@
...
+      field: {
+        borderRadius: '6px',
+        _focus: {
+          boxShadow: '0px 0px 0px 3px rgba(56, 189, 248, 0.6)',
+        },
+        _hover: {
+          border: '1px solid #9CA3AF',
+        },
+        _invalid: {
+          border: '2px solid #EF4444',
```
KonradSzwarc, line 33: 🤔 Similar to the comment above. If there is a need for a thicker border, it would be better to modify only the `borderWidth` property. When it comes to the color you can use `borderColor` property and set its value to one from the theme. This way theme will be the only source of truth when it comes to colors across the app

KonradSzwarc (changes requested):

packages/ui/src/components/NumberInput/NumberInput.tsx
KonradSzwarc: I cannot see the comment it relates to, but to make this error disappear add `defaultValue?: string | number;` to `NumberInputProps` interface

packages/ui/src/theme/overwrites/components/numberInput.ts
talcia: I checked it and it's match by height but not by padding so I delete height from this customize sizes.

hoolek77 (dismissed):

packages/ui/src/components/NumberInput/NumberInput.stories.tsx
```diff
@@ -45,8 +45,8 @@ export const Disabled = () => (
 
 export const Pseudoclasses = () => (
   <HStack>
-    <NumberInput size="md" inputMode="decimal" disabled={false} />
-    <NumberInput size="md" inputMode="numeric" disabled={false} value={1} max={0} />
-    <NumberInput size="md" inputMode="numeric" disabled={false} isFocus />
+    <NumberInput size="md" />
```
KonradSzwarc, line 48: 💭 I think, there is no need to define size when you use the default value for it. Also as you named this story `Pseudoclasses` you should change only props related to them, not the size.

talcia: But in this component size property is required so i need to define size there

KonradSzwarc: So there is the catch - size shouldn't be a required property

KonradSzwarc (approved):

hoolek77 (approved):

KonradSzwarc (approved):

PR #46, opened 2021-07-07
feat(ui): Create radio component (#30)

### LINKED ISSUE #30: Create Radio component
**Acceptance criteria:**
- Uses Radio component from Chakra UI under the hood.
- Its props extend `Omit<HTMLChakraProps<"div">, "onChange" | "defaultChecked">`.
- Receives the following props inherited directly from Chakra UI: value, onChange, children.
- Receives the `checked` prop and passed it as `isChecked`.
- Restricts the original Chakra's radio `size` prop values to `md` and `lg`.

**Mockups:**
For Figma-related reasons, there are two mockups but they should be implemented as one component.
[Radio](https://www.figma.com/file/ur0KLA9ZOtiodrAmpXrxh9/CodersCamp-app-and-website?node-id=180%3A572)
[Radio with label](https://www.figma.com/file/ur0KLA9ZOtiodrAmpXrxh9/CodersCamp-app-and-website?node-id=180%3A617)

kacper-cyra:
Added radio input element with stories and a test.

Closes #30


packages/ui/src/components/Radio/Radio.tsx
```diff
@@ -0,0 +1,45 @@
...
+   */
+  disabled?: boolean;
+  /**
+   * Color variant of the radio
+   */
+  color?: RadioColor;
+  value?: string | number;
+}
+
+export const Radio = forwardRef<RadioProps, 'div'>(
```
KonradSzwarc, line 27: In Chakra `ref` is passed to the `input` element

packages/ui/src/components/Radio/Radio.tsx
```diff
@@ -0,0 +1,45 @@
...
+  ({ color = 'default', value, children = '', checked = false, disabled = false, size = 'md', ...props }, ref) => {
```
KonradSzwarc, line 28: I think there is no need to destructure the `children` property

KonradSzwarc (changes requested):

packages/ui/package.json
```diff
@@ -35,6 +35,7 @@
     "@testing-library/react": "11.2.7",
     "@types/react": "17.0.11",
     "@types/react-dom": "17.0.7",
-    "react": "17.0.2"
+    "react": "17.0.2",
+    "ts-node": "10.0.0"
```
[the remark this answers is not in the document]
kacper-cyra: When I tried to run yarn test, jest displayed error that he cannot compile jest.config.ts without ts-node dependency installed.

packages/ui/src/components/Radio/Radio.tsx
```diff
@@ -0,0 +1,37 @@
+import React from 'react';
+import { forwardRef, HTMLChakraProps, Radio as ChakraRadio } from '@chakra-ui/react';
+
+export type RadioSize = 'md' | 'lg';
+export type RadioColor = 'default' | 'brand';
```
KonradSzwarc, line 5: as unknown as string

KonradSzwarc (dismissed):

packages/ui/package.json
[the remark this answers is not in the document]
KonradSzwarc: Sounds legit, but if that's the case install `ts-node` in the root folder as we use jest in all workspaces

packages/ui/src/components/Radio/Radio.tsx
```diff
@@ -0,0 +1,36 @@
...
+   * Determines if input is disabled.
+   */
+  disabled?: boolean;
+  value: string | number;
+}
+
+export const Radio = forwardRef<RadioProps, 'input'>(
+  ({ value, checked = false, disabled = false, size = 'md', ...props }, ref) => {
+    return (
+      <ChakraRadio
```
KonradSzwarc, line 25: A small thing, but Radio should have a different color when use is clicking it (active state, third row on mockups). You can use [theme component syles](https://chakra-ui.com/docs/theming/component-style), to apply it

packages/ui/src/components/Radio/Radio.tsx
```diff
@@ -0,0 +1,36 @@
...
+   */
+  size?: RadioSize;
+  /**
+   * Determines state of the radio.
+   */
+  checked?: boolean;
+  /**
+   * Determines if input is disabled.
+   */
+  disabled?: boolean;
```
Veryyapeee, line 18: I think we should comment the code only when it's that complicated that it can't speak for itself

packages/ui/src/components/Radio/Radio.tsx
```diff
@@ -0,0 +1,36 @@
...
+  value: string | number;
+}
+
+export const Radio = forwardRef<RadioProps, 'input'>(
+  ({ value, checked = false, disabled = false, size = 'md', ...props }, ref) => {
+    return (
```
Veryyapeee, line 24: "=> (<ChakraRadio />) - simpler version

packages/ui/src/theme/overwrites/components/radio.ts
```diff
@@ -0,0 +1,12 @@
+export const Radio = {
+  baseStyle: {
+    ':active': {
+      border: '6px solid brand.700',
```
KonradSzwarc, line 4: Shouldn't you modify only the `borderColor` property here?

packages/ui/src/theme/overwrites/index.ts
```diff
@@ -1,11 +1,13 @@
 import { extendTheme } from '@chakra-ui/react';
 
+import { components } from './components';
 import { foundations } from './foundations';
 import { styles } from './styles';
 
 export const theme = extendTheme({
   ...foundations,
   styles,
+  components: { ...components },
```
KonradSzwarc, line 10: There is no need to destructure `components` to a new object. You can just pass them as `styles` are passed

packages/ui/src/components/Radio/Radio.tsx
```diff
@@ -0,0 +1,28 @@
+import React from 'react';
+import { forwardRef, HTMLChakraProps, Radio as ChakraRadio } from '@chakra-ui/react';
+
+export type RadioSize = 'md' | 'lg';
```
KonradSzwarc, line 4: There is no need to export this type as we don't use it anywhere else

KonradSzwarc (changes requested):

kacper-cyra:
I added the styling for pseudo classes, but encounter problems with :active. I made it work properly through normal css but in chakra it does not apply at all, no matter what browser.

KonradSzwarc (approved):

hoolek77 (approved):

PR #47, opened 2021-07-09
feat(ui): Create Label component (#34)

### LINKED ISSUE #34: Create Label component
**Acceptance criteria:**
- Uses FormLabel component from Chakra UI under the hood.
- It's props extends `HTMLChakraProps<"label">`.
- Receives the `required` prop as optional boolean. If the prop is `true` displays `*` after the label's content.
- Receives the `children` prop as required `ReactText | ReactText[]`.
- Receives the `size` prop with possible values: `'sm' | 'md' | 'lg'`.

**Mockups:**
[Label](https://www.figma.com/file/ur0KLA9ZOtiodrAmpXrxh9/CodersCamp-app-and-website?node-id=192%3A1304)

lagdush:
Closes #34


packages/ui/src/components/Label/Label.tsx
```diff
@@ -0,0 +1,25 @@
...
+      <FormLabel ref={ref} color="gray.700" fontSize={size} required={required} {...props}>
+        {children} <span style={{ color: 'red' }}>*</span>
+      </FormLabel>
+    );
+  }
+  return (
+    <FormLabel ref={ref} color="gray.700" fontSize={size} required={required} {...props}>
+      {children}
+    </FormLabel>
+  );
```
kacper-cyra, line 24: Maybe we could reduce it to one return. Something like:
`
{children}
{required ? <span style={{ color: 'red' }}>*</span> : ''}
`

kacper-cyra (dismissed):

packages/ui/src/components/Label/Label.tsx
```diff
@@ -0,0 +1,25 @@
...
+  children: ReactText | ReactText[];
+  size?: LabelSizeProps;
+  required?: boolean;
+}
+
+export const Label = forwardRef<LabelProps, 'label'>(({ size = 'sm', required, children, ...props }, ref) => {
+  if (required) {
+    return (
+      <FormLabel ref={ref} color="gray.700" fontSize={size} required={required} {...props}>
+        {children} <span style={{ color: 'red' }}>*</span>
```
KonradSzwarc, line 16: The color shouldn't be just 'red'. It's `red.600` from our theme. To apply it I would suggest using the Box component instead of inline styles. It can look like this `<Box as="span" color="red.600">*</Box>`

packages/ui/src/components/Label/Label.tsx
```diff
@@ -0,0 +1,25 @@
+import React, { ReactText } from 'react';
+import { FormLabel, forwardRef, HTMLChakraProps } from '@chakra-ui/react';
+
+export type LabelSizeProps = 'sm' | 'md' | 'lg';
```
KonradSzwarc, line 4: There is no need to export this type as we don't use it outside of this component

KonradSzwarc (changes requested):

packages/ui/src/components/Label/Label.tsx
```diff
@@ -0,0 +1,25 @@
...
+  children: ReactText | ReactText[];
+  size?: LabelSizeProps;
+  required?: boolean;
+}
+
+export const Label = forwardRef<LabelProps, 'label'>(({ size = 'sm', required, children, ...props }, ref) => {
+  return (
+    <FormLabel ref={ref} color="gray.700" fontSize={size} required={required} {...props}>
+      {children}
+      {required ? (
```
KonradSzwarc, line 16: It can be simplified to `required && <Box as="span" color="red.600">*</Box>`

KonradSzwarc (approved):
One simplification can be added, but overall it looks great 👍

packages/ui/src/components/Label/Label.tsx
lagdush: It's not a problem to add it, but one check from vercel is failed - Vercel-coderscamp-docs.

KonradSzwarc (approved):

packages/ui/src/components/Label/Label.tsx
KonradSzwarc: You can merge the `main` branch - it will fix the problem with Vercel

kacper-cyra (approved):

PR #50, opened 2021-07-10
feat(ui): Create checkbox component (#29)

### LINKED ISSUE #29: Create Checkbox component
**Acceptance criteria:**
- Uses Checkbox component from Chakra UI under the hood.
- It's props extends `Omit<HTMLChakraProps<"div">, keyof UseCheckboxProps>`.
- Receives the following props inherited directly from Chakra UI: all aria-* attributes, value, onChange, name, children.
- Receives the `checked` prop and passed it as `isChecked`.
- Restricts the original Chakra's checkbox `size` prop values to `md` and `lg`.

**Mockups:**
For Figma-related reasons, there are two mockups but they should be implemented as one component.
[Checkbox](https://www.figma.com/file/ur0KLA9ZOtiodrAmpXrxh9/CodersCamp-app-and-website?node-id=40%3A404)
[Checkbox with label](https://www.figma.com/file/ur0KLA9ZOtiodrAmpXrxh9/CodersCamp-app-and-website?node-id=40%3A429)

KrystianKjjk:
- added checkbox component
- added components folder to theme, where we can override default chakra styles 
- updated name convention in esLint, because chakra requires name with underscore at the beginning i.e (_active) 

closes #29


packages/ui/src/components/Checkbox/Checkbox.stories.tsx
```diff
@@ -0,0 +1,66 @@
...
+
+import { HStack, VStack } from '../Stack';
+import { Checkbox, CheckboxProps } from '.';
+
+const UnCheckedCb = (props: Omit<CheckboxProps, 'onChange' | 'checked'>) => (
+  <Checkbox {...props} onChange={() => {}} checked={false} />
+);
+const CheckedCb = (props: Omit<CheckboxProps, 'onChange' | 'checked'>) => (
+  <Checkbox {...props} onChange={() => {}} checked />
+);
```
KonradSzwarc, line 12: I would consider if it's necessary to create such components. The only thing that changes is the `checked` prop and it would be shorter to just pass this property in particular stories (checkbox is unchecked by default)

packages/ui/src/components/Checkbox/Checkbox.stories.tsx
```diff
@@ -0,0 +1,66 @@
...
+Playground.args = {
+  children: 'Checkbox text',
+  size: 'md',
+  disabled: false,
+  checked: false,
+  name: 'terms',
+  value: 'agree',
+};
+
+export const Checkboxes = () => (
```
KonradSzwarc, line 34: This solution looks inconsistent when you look at other components in our codebase. It could be better if you look at the [Radio PR](https://github.com/CodersCrew/coderscamp/pull/46/files) and create checkbox stories in a similar way

packages/ui/src/components/Checkbox/Checkbox.test.tsx
```diff
@@ -0,0 +1,33 @@
...
+  });
+  it('checks that checked checkbox is checked', () => {
+    render(<Checkbox value="test" onChange={() => {}} checked aria-label="checkbox" />);
+
+    const checkbox = screen.getByLabelText('checkbox');
+
+    expect(checkbox).toBeChecked();
+  });
+  it('checks checkbox displays correct labels', () => {
+    const CbLabel = 'checkbox name';
```
KonradSzwarc, line 22: As a good practice, only classes, components, and types should be written CamelCase. It would be better to change it to just `label`

packages/ui/src/components/Checkbox/Checkbox.tsx
```diff
@@ -0,0 +1,31 @@
...
+  disabled?: boolean;
+  size?: 'md' | 'lg';
+}
+
+export const Checkbox = forwardRef<CheckboxProps, 'input'>(
+  ({ checked, disabled = false, size = 'md', ...props }, ref) => {
+    return (
+      <ChakraCheckbox
+        ref={ref}
+        borderRadius={2}
```
KonradSzwarc, line 22: You can specify `borderRadius` and the default `colorScheme` on a theme level. This way we can have single source of truth when it comes to component styles

packages/ui/src/theme/overwrites/components/Checkbox.ts
```diff
@@ -0,0 +1,42 @@
...
+const parts = ['container', 'control', 'label', 'icon'];
+
+function baseStyleControl(props: Record<string, any>) {
+  const { colorScheme: c } = props;
+  const disabledCbColorLight = 'gray.100';
+
+  return {
+    _checked: {
+      _active: {
+        bg: mode(`${c}.700`, `${c}.200`)(props),
```
KonradSzwarc, line 12: I understand that using mode here makes this component more adaptable. Nevertheless, we don't test this component on dark mode anywhere so we cannot be sure if it looks good. In my opinion, it would be better to just pass string values here. We would have an additional gain as there will be no need to use the function to generate styles and that could make code easier to read

packages/ui/src/theme/overwrites/components/Checkbox.ts
```diff
@@ -0,0 +1,42 @@
...
+        borderColor: mode(disabledCbColorLight, 'transparent')(props),
+      },
+    },
+    _hover: {
+      borderColor: mode(`gray.400`, `gray.300`)(props),
+    },
+    _active: {
+      borderColor: mode(`gray.500`, `gray.200`)(props),
+    },
+    _disabled: {
```
KonradSzwarc, line 26: I'm not convinced if we should style disabled hover and active states. Aren't they identical to mockups by default?

packages/ui/src/theme/overwrites/components/Checkbox.ts
```diff
@@ -0,0 +1,42 @@
+import { mode } from '@chakra-ui/theme-tools';
+
+const parts = ['container', 'control', 'label', 'icon'];
```
KonradSzwarc, line 3: There is probably no need to define parts as you only redefine the default ones

packages/ui/src/theme/overwrites/components/Checkbox.ts
```diff
@@ -0,0 +1,42 @@
...
+      },
+    },
+  };
+}
+
+const baseStyle = (props: Record<string, any>) => ({
+  control: baseStyleControl(props),
+});
+
+// eslint-disable-next-line import/no-default-export
```
KonradSzwarc, line 41: I think it could be exported as `export const Checkbox = ...` so there will be no need to disable ESLint

packages/ui/src/theme/overwrites/components/index.ts
```diff
@@ -0,0 +1,6 @@
+import Checkbox from './Checkbox';
+
+// eslint-disable-next-line import/no-default-export
```
KonradSzwarc, line 3: You can write it as `export const components = ...` so ESLint will be happy

KonradSzwarc (changes requested):

.eslintrc.js
```diff
@@ -71,7 +71,7 @@ module.exports = {
     // Enforces some naming conventions across the codebase.
     '@typescript-eslint/naming-convention': [
       2,
-      { leadingUnderscore: 'allow', selector: 'default', format: ['camelCase', 'UPPER_CASE', 'PascalCase'] },
+      { selector: 'default', leadingUnderscore: 'allow', format: ['camelCase', 'UPPER_CASE', 'PascalCase'] },
```
KonradSzwarc: 🙃 Looks like a change that changes nothing

packages/ui/src/components/Checkbox/Checkbox.stories.tsx
```diff
@@ -0,0 +1,54 @@
+import React from 'react';
+import { Meta, Story } from '@storybook/react';
+
+import { HStack, VStack } from '../Stack';
+import { Checkbox, CheckboxProps } from '.';
```
KonradSzwarc, line 5: 💭 It could be better to import it from `./Checkbox`. It lowers our risk of dependency cycles

packages/ui/src/theme/overwrites/components/Checkbox.ts
```diff
@@ -0,0 +1,42 @@
+function baseStyleControl(props: Record<string, unknown>) {
+  const { colorScheme: c } = props;
+  const disabledCbColorLight = 'gray.100';
+
+  return {
+    _checked: {
+      _active: {
+        bg: `${c}.700`,
```
KonradSzwarc, line 8: 💭 It might be enough to write `brand.700` directly. For now, we don't have plans to customize checkbox colors.

packages/ui/src/theme/overwrites/components/Checkbox.ts
KonradSzwarc: I think, as long as we have all styles in one place (theme) it wouldn't be hard to add a dark theme in the future when we will have mockups for it

.eslintrc.js
KrystianKjjk: yea, created while merging, didnt notice, i thought i took incoming changes

KrystianKjjk: That is exactly what i wanted to achieve - srly, thats caused of diff between what i added and what has been added into main earlier

hoolek77 (dismissed):

packages/ui/src/theme/overwrites/components/Checkbox.ts
```diff
@@ -0,0 +1,41 @@
+function baseStyleControl() {
```
KonradSzwarc, line 1: 🤔 Now there is probably no need to use a function here and for `baseStyle`. You can simply create objects.

KonradSzwarc (approved):

hoolek77 (approved):

kamildzierzak (approved):

PR #54, opened 2021-07-11
feat(ui): Create Textarea component (#33)

### LINKED ISSUE #33: Create Textarea component
**Acceptance criteria:**
- Uses Textarea component from Chakra UI under the hood.
- It's props extends `Omit<HTMLChakraProps<"textarea">, 'readonly'>`.
- Receives the following props inherited directly from Chakra UI: value, onChange, name.
- Restricts the original Chakra's textarea `size` prop values to `sm`, `md` and `lg`.

**Mockups:**
[Textarea](https://www.figma.com/file/ur0KLA9ZOtiodrAmpXrxh9/CodersCamp-app-and-website?node-id=183%3A546)

Veryyapeee:
- added textarea component
closes #33


packages/ui/src/theme/overwrites/components/textarea.ts
```diff
@@ -0,0 +1,31 @@
+export const Textarea = {
+  baseStyle: {
+    control: {
+      _disabled: {
```
Veryyapeee, line 4: Idk why when I tried to apply _focus : {boxShadow: 'outline'} it didn't work and I had to put it in the Textarea component itself
Maybe someone else has a clue what I did wrong

kamildzierzak: I had a problem with that when I was creating the Input component. Probably some styles couldn't be set as base, but only as styles for the selected variant. You can try to set styles for outline, e.g .:
```suggestion
export const Textarea = {
  baseStyle: {
    width: '12.5rem',
    height: '5rem',
  },
  sizes: {
    sm: {
      px: '0.75rem',
      py: '0.375rem',
    },
    md: {
      px: '1rem',
      py: '0.5rem',
    },
    lg: {
      px: '1rem',
      py: '0.625rem',
    },
    defaultProps: {
      size: 'md',
    },
  },
  variants: {
    outline: {
      _disabled: {
        opacity: '0.4',
      },
      _focus: {
        boxShadow: 'outline',
      },
      _hover: {
        borderColor: 'gray.400',
      },
    },
  },
};

```

Veryyapeee: Great, thanks!

kamildzierzak (approved):

packages/ui/src/components/Textarea/Textarea.tsx
```diff
@@ -0,0 +1,20 @@
+import React, { ChangeEventHandler } from 'react';
+import { forwardRef, HTMLChakraProps, Textarea as ChakraTextarea } from '@chakra-ui/react';
+
+type TextareaSize = 'sm' | 'md' | 'lg';
+
+export interface TextareaProps extends Omit<HTMLChakraProps<'textarea'>, 'readonly'> {
+  size?: TextareaSize;
+  disabled?: boolean;
+  onChange?: ChangeEventHandler<HTMLTextAreaElement> | undefined;
```
KonradSzwarc, line 9: Is there a need for `| undefined` for the optional property?

packages/ui/src/components/Textarea/index.ts
```diff
@@ -0,0 +1 @@
+export * from './Textarea';
```
KonradSzwarc, line 1: To make codebase more consistent it would be great to export TextArea and its props separately

packages/ui/src/theme/overwrites/components/textarea.ts
```diff
@@ -0,0 +1,37 @@
+export const Textarea = {
+  baseStyle: {
```
KonradSzwarc, line 2: I think we shouldn't define the default width and height for the Textarea. We would probably never need it exactly in this size so size should be defined only for a particular use cases

packages/ui/src/theme/overwrites/components/textarea.ts
```diff
@@ -0,0 +1,37 @@
...
+      px: '1rem',
+      py: '0.5rem',
+    },
+    lg: {
+      px: '1rem',
+      py: '0.625rem',
+    },
+    defaultProps: {
+      size: 'md',
+    },
```
KonradSzwarc, line 35: `md` is the default value for Textarea, so I think there is no need do redefine it again

packages/ui/src/theme/overwrites/components/textarea.ts
```diff
@@ -0,0 +1,37 @@
+export const Textarea = {
+  baseStyle: {
+    width: '12.5rem',
+    height: '5rem',
+  },
+  variants: {
+    outline: {
+      borderRadius: '0.375rem',
```
KonradSzwarc, line 8: It would be great to provide all sizes non-related to fonts in pixels. This way we don't to recalculate it every time we want to check if the component match mockups

packages/ui/src/theme/overwrites/components/textarea.ts
```diff
@@ -0,0 +1,37 @@
...
+      _focus: {
+        boxShadow: 'outline',
+      },
+      _disabled: {
+        opacity: '0.4',
+      },
```
KonradSzwarc, line 14: That's the default value so we probably don't need to repeat it here

packages/ui/src/components/Textarea/Textarea.stories.tsx
```diff
@@ -0,0 +1,36 @@
...
+};
+
+export default meta;
+
+const Template: Story<TextareaProps> = (args) => <Textarea {...args} />;
+
+export const Playground = Template.bind({});
+
+Playground.args = {
+  placeholder: 'Default textarea',
```
KonradSzwarc, line 19: It would be great to provide values for the size, disabled, value, name, and invalid. This way Storybook will initially display correct controls for them.

packages/ui/src/theme/overwrites/components/textarea.ts
```diff
@@ -0,0 +1,37 @@
+export const Textarea = {
+  baseStyle: {
+    width: '12.5rem',
+    height: '5rem',
+  },
+  variants: {
+    outline: {
```
KonradSzwarc, line 7: I think we should specify the `gray.300` as border color because by default its value is `gray.200`

KonradSzwarc (changes requested):

hoolek77 (approved):

packages/ui/src/theme/overwrites/components/textarea.ts
```diff
@@ -0,0 +1,34 @@
...
+    md: {
+      px: '16px',
+      py: '8px',
+    },
+    lg: {
+      px: '16px',
+      py: '10px',
+    },
+  },
+};
```
hoolek77, line 34: So, there's a problem with this file, styles aren't applied correctly and there are typing problems. I've checked it locally and found a solution for that. You should wrap those properties in a field object. For example: 
```typescript
baseStyle: {
    field: {
      borderColor: 'gray.300',
    }
  }
```
You can also take a look at how was it implemented in radio overwrites because it's very similar.

hoolek77 (changes requested):

KonradSzwarc (approved):

hoolek77 (approved):

KonradSzwarc (approved):

PR #55, opened 2021-07-11
feat(ui): Create Input component (#31)

### LINKED ISSUE #31: Create Input component
**Acceptance criteria:**
- Uses Input component from Chakra UI under the hood.
- It's props extends `Omit<HTMLChakraProps<"input">, 'readonly' | 'size'>`.
- Receives the following props inherited directly from Chakra UI: value, onChange, name.
- Restricts the original Chakra's input `size` prop values to `sm`, `md` and `lg`.

**Mockups:**
[Input](https://www.figma.com/file/ur0KLA9ZOtiodrAmpXrxh9/CodersCamp-app-and-website?node-id=41%3A417)

kamildzierzak:
Closes #31

### DIFF
--- a/packages/ui/src/components/Input/Input.stories.tsx
+++ b/packages/ui/src/components/Input/Input.stories.tsx
@@ -24,14 +24,14 @@ Playground.args = {
 export const Sizes = () => (
   <HStack>
     <Input placeholder="Small (sm)" size="sm" />
-    <Input placeholder="Medium (md)" size="md" />
+    <Input placeholder="Medium (md)" />
     <Input placeholder="Large (lg)" size="lg" />
   </HStack>
 );
 
 export const OtherStates = () => (
   <HStack>
-    <Input isInvalid="true" placeholder="Invalid!" size="md" />
-    <Input disabled placeholder="Disabled" size="md" />
+    <Input isInvalid="true" placeholder="Invalid!" />
+    <Input disabled placeholder="Disabled" />
   </HStack>
 );
--- a/packages/ui/src/components/Input/Input.stories.tsx
+++ b/packages/ui/src/components/Input/Input.stories.tsx
@@ -32,5 +32,6 @@ export const Sizes = () => (
 export const OtherStates = () => (
   <HStack>
     <Input isInvalid="true" placeholder="Invalid!" size="md" />
+    <Input disabled placeholder="Disabled" size="md" />
   </HStack>
 );
--- a/packages/ui/src/components/Input/Input.stories.tsx
+++ b/packages/ui/src/components/Input/Input.stories.tsx
@@ -1,0 +1,36 @@
+import React from 'react';
+import { Meta, Story } from '@storybook/react';
+
+import { HStack } from '../Stack';
+import { Input, InputProps } from './Input';
+
+const meta: Meta = {
+  title: 'Components / Input',
+  component: Input,
+};
+
+export default meta;
+
+const Template: Story<InputProps> = (args) => <Input {...args} />;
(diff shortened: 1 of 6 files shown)


packages/ui/src/components/Input/Input.stories.tsx
```diff
@@ -0,0 +1,37 @@
...
+  <HStack>
+    <Input placeholder="Small (sm)" size="sm" />
+    <Input placeholder="Medium (md)" size="md" />
+    <Input placeholder="Large (lg)" size="lg" />
+  </HStack>
+);
+
+export const OtherStates = () => (
+  <HStack>
+    <Input isInvalid="true" placeholder="Invalid!" size="md" />
```
KonradSzwarc, line 34: I think there is no need to provide props that don't affect states you want to show, like `size` in this case. Especially in cases like here when `md` is the default value anyway.

packages/ui/src/components/Input/Input.test.tsx
```diff
@@ -0,0 +1,22 @@
+import React from 'react';
+import { render, screen } from '@testing-library/react';
+
+import { Input } from './index';
```
KonradSzwarc, line 4: To reduce the chance for import cycles you can import Input from the `./Input` file

packages/ui/src/theme/overwrites/components/input.ts
```diff
@@ -0,0 +1,32 @@
+export const Input = {
+  sizes: {
+    sm: {
+      field: {
+        padding: '0.375rem 0.75rem',
```
KonradSzwarc: It can be better to use pixels for anything non-related to fonts. This way it would be easier for us to quickly check if values match mockups.

packages/ui/src/theme/overwrites/components/input.ts
```diff
@@ -0,0 +1,32 @@
...
+        padding: '10px 16px',
+        borderRadius: '6px',
+      },
+    },
+  },
+  variants: {
+    outline: {
+      field: {
+        _focus: {
+          background: '#FFF',
```
KonradSzwarc, line 26: We shouldn't use colors that aren't included in the theme. Here you can safely write `white` as Chakra has such value in its default theme.

packages/ui/src/theme/overwrites/components/input.ts
```diff
@@ -0,0 +1,32 @@
+export const Input = {
+  sizes: {
+    sm: {
+      field: {
+        padding: '6px 12px',
+        borderRadius: '6px',
```
KonradSzwarc, line 6: I think you can define `borderRadius` in the outline variant config to don't repeat yourself

packages/ui/src/theme/overwrites/components/input.ts
```diff
@@ -0,0 +1,32 @@
...
+    lg: {
+      field: {
+        padding: '10px 16px',
+        borderRadius: '6px',
+      },
+    },
+  },
+  variants: {
+    outline: {
+      field: {
```
KonradSzwarc, line 24: You should probably specify border colors for different states as Chakra's defaults are different than our mockups

KonradSzwarc (changes requested):

packages/ui/src/theme/overwrites/components/input.ts
kamildzierzak: Yes, indeed 😅

KonradSzwarc (approved):

hoolek77 (approved):

PR #59, opened 2021-07-17
feat(ui): Radio group component (#37)

### LINKED ISSUE #37: Create RadioGroup component
**Acceptance criteria:**
- Uses RadioGroup component from Chakra UI under the hood.
- Wraps children with the Stack component from Chakra UI.
- Its interface extends StackProps from Chakra UI.
- Receives the following props inherited directly from Chakra UI: value, onChange, name, children.
- Restricts the original Chakra's RadioGroup `size` prop values to `md` and `lg`.

hoolek77:
I didn't add a disabled prop because I've checked that it doesn't affect the ChakraRadioGroup component at all. 
Also, I've had to modify the Radio component, so it doesn't assign size prop, because RadioGroup size wasn't changing Radio sizes.

Closes #37

### DIFF
--- a/packages/ui/src/components/RadioGroup/RadioGroup.test.tsx
+++ b/packages/ui/src/components/RadioGroup/RadioGroup.test.tsx
@@ -1,0 +1,33 @@
+import React from 'react';
+import { render, screen } from '@testing-library/react';
+
+import { Radio } from '../Radio';
+import { RadioGroup } from './RadioGroup';
+
+describe('Radio', () => {
+  const textOne = 'One';
+  const textTwo = 'Two';
+
+  it('renders correctly', () => {
+    render(<RadioGroup name="test" value="1" onChange={() => {}} data-testid="radio" />);
+
+    const radio = screen.getByTestId('radio');
+
+    expect(radio).toBeInTheDocument();
+  });
+
+  it('renders correctly checked radio', () => {
+    render(
+      <RadioGroup name="test" value="1" onChange={() => {}}>
+        <Radio value="1">{textOne}</Radio>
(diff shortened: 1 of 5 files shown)


packages/ui/src/components/RadioGroup/RadioGroup.test.tsx
```diff
@@ -0,0 +1,33 @@
+import React from 'react';
+import { render, screen } from '@testing-library/react';
+
+import { Radio } from '../Radio';
+import { RadioGroup } from './RadioGroup';
+
+describe('Radio', () => {
```
KonradSzwarc, line 7: 🤔 It should probably be 'RadioGroup', not 'Radio'

packages/ui/src/components/RadioGroup/RadioGroup.tsx
```diff
@@ -0,0 +1,19 @@
+import React from 'react';
+import { RadioGroup as ChakraRadioGroup, RadioGroupProps as ChakraRadioGroupProps } from '@chakra-ui/react';
+
+import { Stack, StackProps } from '../Stack';
+
+type RadioGroupSize = 'md' | 'lg';
+
+export interface RadioGroupProps extends Omit<StackProps, 'onChange'> {
+  size?: RadioGroupSize;
```
KonradSzwarc, line 9: 💭 Maybe we can import `RadioSize` from the `Radio` component and use it here. This way it would be always guaranteed that RadioGroup handles all possible sizes of the `Radio`.

KonradSzwarc (approved):

talcia (approved):

PR #74, opened 2021-07-20
ci: fix tests in CI (#73)

### LINKED ISSUE #73: Fix tests in CI
Currently, our `test` phase in CI always finishes without running any tests. As it's important for us to fix it fast, the scope of this task is to run tests with all files. Running tests only for affected files will be in the scope of a separate task.

KonradSzwarc:
Closes #73


hoolek77 (approved):

kamildzierzak (approved):

mecies (approved):
Out of curiosity, is something wrong with the onlyChanged script? Why is the change required? 🙏

KonradSzwarc:
> Out of curiosity, is something wrong with the onlyChanged script? Why is the change required? 🙏

Yes, it doesn't work 😅 You can check one of the PRs with new components. They always have some tests but none of them run. This PR provides the quickest solution, but of course, in the future, we need to move back to this topic. @mecies I remember you set up such scripts for tests and linter in the previous project. Maybe you want to write create them for this project as well?

PR #76, opened 2021-07-22
feat(ui): create ProjectCard component (#63)

### LINKED ISSUE #63: Create ProjectCard component
**Acceptance criteria:**
- Has 100% width by default.
- Receives the required `status` prop with one of values: `idle`, `doing`, `review`, `done`.
- Receives the required `image`, `title`, and `url` props of type string.
- When `status` is different than `done` receives the required `date` prop of type Date.
- When `status` is `done` receives the required `points` and `pointsMax` props of type number.

**Mockups:**
[Figma link](https://www.figma.com/file/ur0KLA9ZOtiodrAmpXrxh9/CodersCamp-app-and-website?node-id=630%3A1674)

lagdush:
closes #63


packages/ui/src/components/ProjectCard/ProjectCard.tsx
```diff
@@ -0,0 +1,104 @@
...
+      statusText = 'W ocenie';
+      timeStatus = 'Demo';
+      dateOrPointsText = dateProps.toLocaleDateString();
+    }
+    if (status === 'done') {
+      statusText = 'Zakończony';
+      timeStatus = 'Punkty';
+      dateOrPointsText = `${points}/${pointsMax}`;
+    }
+
```
Veryyapeee, line 69: 🏕️ So in case of performance we should replace these if statements with IMO switch case, or if you don't want to, use else if, this way performance is better, right now it's going through and checking every if statement even if previous check passed
I'm also a fan of extracting such a helpers outside of the component, but it's up to you :D
And also, for future purposes I think we should make an enum for "status" props and export it, I don't think that using in future plain strings as a props is good and maintainable idea

packages/ui/src/components/ProjectCard/ProjectCard.stories.tsx
```diff
@@ -0,0 +1,52 @@
...
+  pointsMax: 200,
+};
+
+export const Variants = () => (
+  <SimpleGrid columns={4} spacing={3} h="444px" w="320">
+    <ProjectCard status="idle" image="test" title={title} url="test" date={date} />
+    <ProjectCard status="doing" image="test" title={title} url="test" date={date} />
+    <ProjectCard status="review" image="test" title={title} url="test" date={date} />
+    <ProjectCard status="done" image="test" title={title} url="test" points={100} pointsMax={200} />
+  </SimpleGrid>
```
Veryyapeee, line 51: 🏕️Same as said in previous comment about enums for status (or constants)
The other thing is you declared the object with values for status props and you still used plain strings in here, I think we should refactor that :D

packages/ui/src/components/ProjectCard/ProjectCard.tsx
```diff
@@ -0,0 +1,104 @@
...
+    return (
+      <Box status={status} ref={ref} w="100%" boxShadow="base" borderRadius="8px" {...props}>
+        <Grid templateRows="repeat(2, 1fr)" h="100%" gap="24px">
+          <Box as="img" src={image} alt="" background="gray.300" borderRadius="8px 8px 0 0" h="auto" />
+
```
Veryyapeee, line 74: 🏕️ I think we can extract these gap, borderRadius etc. values to theme, this way we could use them in other card components etc. :)

packages/ui/src/components/ProjectCard/ProjectCard.tsx
```diff
@@ -0,0 +1,104 @@
...
+  | {
+      status: 'idle' | 'doing' | 'review';
+      date: Date;
+    }
+  | {
+      status: 'done';
+      points: number;
+      pointsMax: number;
+    }
+>;
```
KonradSzwarc, line 36: 🤔 I think this could be achieved easier if you create a type like below
```ts
type DoneProps = {
    status: 'done';
    points: number;
    pointsMax: number;
}

type NotDoneProps = {
    status: 'idle' | 'doing' | 'review';
    date: Date;
}

export type ProjectCardProps = {
    image: string;
    title: string;
    url: string;
}  & (DoneProps | NotDoneProps)
```

packages/ui/src/components/ProjectCard/ProjectCard.tsx
KonradSzwarc: I agree it would be better to move this to a separate function outside of the component. In such a case you can as well use if's with early returns.

When it comes to the enums idea - we have enums usage listed as a bad practice in our docs. Of course @Veryyapeee  if you want to discuss this topic we can do it.

KonradSzwarc (changes requested):

packages/ui/src/components/ProjectCard/ProjectCard.tsx
lagdush: I think you have right. I'll change it.

packages/ui/src/components/ProjectCard/ProjectCard.tsx
lagdush: As Konrad noted, we do not use enums in this project. I will leave status type as it is.

packages/ui/src/components/ProjectCard/ProjectCard.stories.tsx
```diff
@@ -0,0 +1,50 @@
...
+  <Box h="444px" w="320px">
+    <ProjectCard {...args} />
+  </Box>
+);
+
+export const Playground = Template.bind({});
+
+Playground.args = {
+  status: 'idle',
+  image: 'test',
```
KonradSzwarc, line 37: 💭 You can provide some real image URL (e.g. from Unsplash) here and in the variants section

packages/ui/src/components/ProjectCard/ProjectCard.test.tsx
```diff
@@ -0,0 +1,42 @@
...
+const date = new Date();
+const statusArray: StatusArray = [
+  { status: 'idle', text: 'Nie rozpoczęty' },
+  { status: 'doing', text: 'W trakcie' },
+  { status: 'review', text: 'W ocenie' },
+  { status: 'done', text: 'Zakończony' },
+];
+
+describe('ProjectCard', () => {
+  statusArray.map((el) => {
```
KonradSzwarc, line 21: 💭 I think, creating four separate tests instead of array mapping can be easier to read as there would be no conditional logic to track in the test.

packages/ui/src/components/ProjectCard/ProjectCard.test.tsx
```diff
@@ -0,0 +1,42 @@
...
+    return it(`if status is equal to${el.status} should display ${el.text}`, () => {
+      if (el.status === 'done') {
+        render(<ProjectCard status="done" image="test" title={title} url="test" points={100} pointsMax={200} />);
+      } else {
+        render(<ProjectCard status={el.status} image="test" title={title} url="test" date={date} />);
+      }
+
+      expect(screen.getByText(el.text)).toBeInTheDocument();
+    });
+  });
```
KonradSzwarc, line 31: 🙃 Add a new line between different test cases. It will make the code more consistent.

packages/ui/src/components/ProjectCard/ProjectCard.tsx
```diff
@@ -0,0 +1,75 @@
...
+
+import { statusValueCheck } from '../../helpers/statusValueCheck';
+import { Button } from '../Button/Button';
+import { Typography } from '../Typography/Typography';
+
+type DoneProps = {
+  status: 'done';
+  points: number;
+  pointsMax: number;
+  date?: never;
```
KonradSzwarc, line 12: 🤔 Instead of defining prop with type `never` you can simply don't add this prop at all.

packages/ui/src/helpers/statusValueCheck.ts
```diff
@@ -0,0 +1,47 @@
+type StatusProps = 'idle' | 'doing' | 'review' | 'done';
+type StatusCase = {
+  statusText: string;
+  timeStatus: string;
+  dateOrPoints: string;
+  isDoing: boolean;
+};
+
+export const statusValueCheck = (status: StatusProps, points: number, pointsMax: number, dateProps: Date) => {
```
KonradSzwarc, line 9: 🤔 There is probably no need to create a separate helpers directory for this logic, because:
1. The function will be used only in the ProjectCard component.
2. This way you need to define some types twice, which makes code less DRY.

I think it would be better to move this function back to the component file.

packages/ui/src/helpers/statusValueCheck.ts
```diff
@@ -0,0 +1,47 @@
...
+  let statusCase: StatusCase;
+
+  switch (status) {
```
KonradSzwarc, line 12: 🤔 To make it more readable you can use early returns instead of the `switch` and value overwrites.

packages/ui/src/components/ProjectCard/ProjectCard.tsx
```diff
@@ -0,0 +1,75 @@
...
+};
+
+type NotDoneProps = {
+  status: 'idle' | 'doing' | 'review';
+  date: Date;
+  points?: never;
+  pointsMax?: never;
```
KonradSzwarc, line 19: 🤔 Instead of defining prop with type `never` you can simply don't add this prop at all.

packages/ui/src/components/ProjectCard/ProjectCard.tsx
```diff
@@ -0,0 +1,75 @@
...
+export const ProjectCard = forwardRef<ProjectCardProps, 'div'>(
+  // disabled for url props
+  // eslint-disable-next-line @typescript-eslint/no-unused-vars
+  ({ status, image, title, url, points, pointsMax, date, ...props }, ref) => {
+    const { statusText, timeStatus, dateOrPoints, isDoing } = statusValueCheck(
+      status,
+      points as number,
+      pointsMax as number,
+      date as Date,
+    );
```
KonradSzwarc, line 39: 🤔 You can pass one argument to the function and type it as `DoneProps | NotDoneProps`. This way there would be no need for casting.

packages/ui/src/components/ProjectCard/ProjectCard.tsx
```diff
@@ -0,0 +1,75 @@
...
+  image: string;
+  title: string;
+  url: string;
+};
+
+export type ProjectCardProps = CommonProps & (DoneProps | NotDoneProps);
+
+export const ProjectCard = forwardRef<ProjectCardProps, 'div'>(
+  // disabled for url props
+  // eslint-disable-next-line @typescript-eslint/no-unused-vars
```
KonradSzwarc, line 32: 🤔 Why URL is not used?

KonradSzwarc (changes requested):

packages/ui/src/components/ProjectCard/ProjectCard.stories.tsx
```diff
@@ -0,0 +1,50 @@
...
+import { ProjectCard, ProjectCardProps } from './ProjectCard';
+
+const title = 'Zaawansowany React i Node';
+const date = new Date();
+
+const status = { done: 'done', idle: 'idle', doing: 'doing', review: 'review' };
+const meta: Meta = {
+  title: 'Components / ProjectCard',
+  component: ProjectCard,
+  argTypes: {
```
KonradSzwarc, line 14: 🤔 You should add the `date` prop here and type it correctly, as now Storybook displays it as a string in the controls panel.

packages/ui/src/components/ProjectCard/ProjectCard.tsx
```diff
@@ -0,0 +1,75 @@
...
+
+type CommonProps = {
+  image: string;
+  title: string;
+  url: string;
+};
+
+export type ProjectCardProps = CommonProps & (DoneProps | NotDoneProps);
+
+export const ProjectCard = forwardRef<ProjectCardProps, 'div'>(
```
KonradSzwarc, line 30: 🤔 I looked on the card in the Storybook and it doesn't seem right. The content of the card should fill the entire space instead of being centered.
CleanShot 2021-07-25 at 10 36 55@2x

packages/ui/src/components/ProjectCard/ProjectCard.tsx
```diff
@@ -0,0 +1,75 @@
...
+            <Flex justifyContent="space-between" alignSelf="start">
+              <Typography lineHeight="md" fontWeight="500" as="span">
+                {timeStatus}:
+              </Typography>
+              <Typography lineHeight="md" as="span">
+                {dateOrPoints}
+              </Typography>
+            </Flex>
+
+            {isDoing ? <Button>Przejdź do projektu</Button> : <Button disabled>Przejdź do projektu</Button>}
```
KonradSzwarc, line 69: 💭 You probably don't need this `isDoing` helper variable nor the entire ternary. It could be written simpler as
```
<Button disabled={status === 'idle'}>Przejdź do projektu</Button>
```

packages/ui/src/components/ProjectCard/ProjectCard.tsx
```diff
@@ -0,0 +1,75 @@
...
+              <Typography lineHeight="md" as="span">
+                {statusText}
+              </Typography>
+            </Flex>
+
+            <Flex justifyContent="space-between" alignSelf="start">
+              <Typography lineHeight="md" fontWeight="500" as="span">
+                {timeStatus}:
+              </Typography>
+              <Typography lineHeight="md" as="span">
```
KonradSzwarc, line 64: 🏕️ I see you found a bug with the Typography component. Instead of providing `lineHeight` here, I would suggest adding `lineHeight={size} letterSpacing={size}` to the Typography component itself.

packages/ui/src/components/ProjectCard/ProjectCard.tsx
```diff
@@ -0,0 +1,75 @@
+import React from 'react';
+import { Box, Flex, forwardRef, Grid } from '@chakra-ui/react';
+
+import { statusValueCheck } from '../../helpers/statusValueCheck';
+import { Button } from '../Button/Button';
```
KonradSzwarc, line 5: 🙃 It should be enough to import it from '../Button' (similarly in the case of Typography)

packages/ui/src/components/ProjectCard/ProjectCard.tsx
```diff
@@ -0,0 +1,75 @@
...
+  ({ status, image, title, url, points, pointsMax, date, ...props }, ref) => {
+    const { statusText, timeStatus, dateOrPoints, isDoing } = statusValueCheck(
+      status,
+      points as number,
+      pointsMax as number,
+      date as Date,
+    );
+
+    return (
+      <Box status={status} ref={ref} w="100%" h="100%" boxShadow="base" borderRadius="8px" {...props}>
```
KonradSzwarc, line 42: 🤔 Why Box receives the `status` prop. Can this component handle it?

packages/ui/src/components/ProjectCard/ProjectCard.tsx
```diff
@@ -0,0 +1,75 @@
...
+        <Grid templateRows="repeat(2, 1fr)" h="100%" gap="24px">
+          <Box as="img" src={image} alt="" background="gray.300" borderRadius="8px 8px 0 0" h="auto" />
```
KonradSzwarc, line 44: 🤔 It would be better to use Chakra's Image component here. You can also provide something like `Grafika projektu z modułu ${title}` as the `alt` prop.

KonradSzwarc (changes requested):

packages/ui/src/components/ProjectCard/ProjectCard.tsx
lagdush: Adding type never here makes the type safer. This means the component will accept either props date or props points and pointsMax. If we do not add the type never here, Typescript will ask for the appropriate props, but will also accept the extra ones. It will not show an error if I give the status idle, I will give props date and props points.
Zrzut ekranu 2021-07-25 o 16 03 59

packages/ui/package.json
```diff
@@ -20,7 +20,10 @@
     "@chakra-ui/theme-tools": "1.1.8",
     "@emotion/react": "11.4.0",
     "@emotion/styled": "11.3.0",
-    "framer-motion": "4.1.17"
+    "@types/react-router-dom": "5.1.8",
+    "framer-motion": "4.1.17",
+    "moment": "2.29.1",
```
KonradSzwarc, line 25: 🤔 I'm not sure if we should add another project dependency just to display a date in a particular format. Even in case, we would need some library in the future it shouldn't be moment.js. You can read more about it [on their official website](https://momentjs.com/docs/#/-project-status/).

packages/ui/src/components/ProjectCard/ProjectCard.tsx
```diff
@@ -1,75 +1,55 @@
...
+import { useHistory } from 'react-router-dom';
+import { Box, Flex, Grid, Image } from '@chakra-ui/react';
+
+import { Button } from '../Button';
+import { Typography } from '../Typography';
+import { DoneProps, NotDoneProps, ProjectCardProps } from './projectCardTypes';
+import { statusValueCheck } from './statusValueCheck';
+
+export const ProjectCard = ({ status, image, title, url, points, pointsMax, date, ...props }: ProjectCardProps) => {
+  const history = useHistory();
```
KonradSzwarc, line 11: 🤔 ui components shouldn't rely on any specific routing library. We want to use them both in the panel and website (where we have Next.js). You can add `onButtonClick` prop to the component, so we can customize the logic of the function when implementing this component

packages/ui/src/components/ProjectCard/ProjectCard.tsx
```diff
@@ -1,75 +1,55 @@
...
+
+  const { statusText, timeStatus, dateOrPoints } = statusValueCheck(
+    ({ status, points, pointsMax } as DoneProps) || ({ status, date } as NotDoneProps),
```
KonradSzwarc, line 14: 🤔 We shouldn't cast values as it makes our code less type-safe

KonradSzwarc (changes requested):

PR #78, opened 2021-07-25
feat(ui): TeamCard component (#62)

### LINKED ISSUE #62: Create TeamCard component
**Acceptance criteria:**
- Has 100% width by default.
- Receives the `active` prop of a boolean type.
- Receives the `members` prop with `TeamMember[]` type.
- Receives the `mentors` prop with `TeamMember[]` type.
- Uses the `Avatar` component to render members.
- When no avatar provided renders initials derived from the `name` prop of the member.
- When one of the avatars is clicked it should redirect to the `profileUrl`.

**Interfaces:**
```ts
interface TeamMember {
    id: number;
    name: string;
    image?: string;
    profileUrl: string;
}
```

**Mockups:**
[Figma link](https://www.figma.com/file/ur0KLA9ZOtiodrAmpXrxh9/CodersCamp-app-and-website?node-id=634%3A1526)

hoolek77:
Closes #62


packages/ui/src/components/Avatar.tsx
```diff
@@ -0,0 +1,2 @@
+export { Avatar, AvatarGroup } from '@chakra-ui/react';
+export type { AvatarProps } from '@chakra-ui/react';
```
KonradSzwarc, line 2: 👍 Great idea to re-export these components instead of importing them from Chakra in the TeamCard file.

KonradSzwarc (changes requested):

packages/ui/src/components/TeamCard/teamMembers.ts
```diff
@@ -0,0 +1,49 @@
+import { TeamMember } from './TeamCard';
+
+const imageUrl =
```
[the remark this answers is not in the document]
KonradSzwarc: As an alternative, you can leave it this way, but rename the file to TeamCard.mocks.ts to clearly communicate its purpose. Choose the option that seems better for you or feel free to disagree.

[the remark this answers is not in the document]
hoolek77: I think the second option fits well with our file naming conventions, so I've chosen that.

packages/ui/src/components/TeamCard/TeamCard.tsx
```diff
@@ -19,22 +22,23 @@ export interface TeamCardProps extends BoxProps {
   active: boolean;
 }
 
-export const notActiveText =
-  'Jeszcze nie działasz w żdanym zespole. Zostaniesz do niego przydzielony automatycznie po napisaniu pierwszego testu i zakwalifikowaniu się na kurs.';
+type AvatarProps = {
```
KonradSzwarc, line 25: 🙃 `AvatarsProps` probably

packages/ui/src/theme/overwrites/components/link.ts
```diff
@@ -0,0 +1,7 @@
+export const Link = {
+  baseStyle: {
+    _hover: {
+      textDecoration: 'none',
```
KonradSzwarc, line 4: 💭 In many cases we probably want links to have an underline. It could be better to remove this Link-related styling directly in the TeamCard component.

KonradSzwarc (approved):

mecies (approved):
🏎️

PR #86, opened 2021-07-27
feat(ui): Create MaterialsCard Component (#61)

### LINKED ISSUE #61: Create MaterialsCard component
**Acceptance criteria:**
- Has 100% width by default.
- Receives the `status` prop with values: `idle`, `loading` and `generated`.

**Mockups:**
[Figma link](https://www.figma.com/file/ur0KLA9ZOtiodrAmpXrxh9/CodersCamp-app-and-website?node-id=634%3A1556)

talcia:
Closes #61

### DIFF
--- a/packages/ui/src/components/MaterialsCard/MaterialsCard.test.tsx
+++ b/packages/ui/src/components/MaterialsCard/MaterialsCard.test.tsx
@@ -1,0 +1,28 @@
+import React from 'react';
+import { render, screen } from '@testing-library/react';
+
+import { MaterialsCard } from './MaterialsCard';
+
+describe('MaterialsCard', () => {
+  it('should renders correctly in idle status', () => {
+    render(<MaterialsCard />);
+
+    const button = screen.getByText('Wygeneruj listę');
+    expect(button).toBeEnabled();
+    expect(button).toBeInTheDocument();
+  });
+  it('should renders correctly in loading status', () => {
+    render(<MaterialsCard status="loading" />);
+
+    const button = screen.getByText('Generowanie listy...');
+    expect(button).toBeDisabled();
+    expect(button).toBeInTheDocument();
+  });
+  it('should renders correctly in generated status', () => {
+    render(<MaterialsCard status="generated" />);
+
+    const button = screen.getByText('Przejdź do listy materiałów');
+    expect(button).toBeEnabled();
+    expect(button).toBeInTheDocument();
+  });
+});
--- a/packages/ui/src/components/MaterialsCard/MaterialsCard.test.tsx
+++ b/packages/ui/src/components/MaterialsCard/MaterialsCard.test.tsx
@@ -1,13 +1,13 @@
 import React from 'react';
 import { render, screen } from '@testing-library/react';
 
-import { MaterialsCard } from './MaterialsCard';
+import { MaterialsCard, MaterialsCardText } from './MaterialsCard';
 
 describe('MaterialsCard', () => {
   it('should renders correctly in idle status', () => {
     render(<MaterialsCard status="idle" />);
 
-    const button = screen.getByText('Wygeneruj listę');
+    const button = screen.getByText(MaterialsCardText.idle.buttonText);
     expect(button).toBeEnabled();
 
     expect(button).toBeInTheDocument();
@@ -15,7 +15,7 @@ describe('MaterialsCard', () => {
   it('should renders correctly in loading status', () => {
     render(<MaterialsCard status="loading" />);
 
-    const button = screen.getByText('Generowanie listy...');
+    const button = screen.getByText(MaterialsCardText.loading.buttonText);
     expect(button).toBeDisabled();
 
     expect(button).toBeInTheDocument();
@@ -23,7 +23,7 @@ describe('MaterialsCard', () => {
   it('should renders correctly in generated status', () => {
     render(<MaterialsCard status="generated" />);
 
-    const button = screen.getByText('Przejdź do listy materiałów');
+    const button = screen.getByText(MaterialsCardText.generated.buttonText);
     expect(button).toBeEnabled();
 
     expect(button).toBeInTheDocument();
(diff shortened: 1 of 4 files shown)


KonradSzwarc:
When creating a PR remember to add "Closes #ISSUE_ID" to its description

packages/ui/src/components/MaterialsCard/MaterialsCard.test.tsx
```diff
@@ -0,0 +1,28 @@
...
+import { MaterialsCard } from './MaterialsCard';
+
+describe('MaterialsCard', () => {
+  it('should renders correctly in idle status', () => {
+    render(<MaterialsCard />);
+
+    const button = screen.getByText('Wygeneruj listę');
+    expect(button).toBeEnabled();
+    expect(button).toBeInTheDocument();
+  });
```
KonradSzwarc, line 13: 🙃 For consistency, add new lines between tests

packages/ui/src/components/MaterialsCard/MaterialsCard.tsx
```diff
@@ -0,0 +1,53 @@
...
+import { Button } from '../Button';
+import { Typography } from '../Typography/Typography';
+
+type MaterialsCardStatus = 'idle' | 'loading' | 'generated';
+
+export interface MaterialsCardProps extends BoxProps {
+  status: MaterialsCardStatus;
+}
+
+type MaterialsCardOptions = { [status: string]: { statusText: string; buttonText: string } };
```
KonradSzwarc, line 13: 💭 You can type it as `Record<MaterialsCardStatus, { statusText: string; buttonText: string }>`

packages/ui/src/components/MaterialsCard/MaterialsCard.tsx
```diff
@@ -0,0 +1,53 @@
...
+  },
+  generated: {
+    statusText: 'Twoja lista materiałów znajduje się na platformie Process.st',
+    buttonText: 'Przejdź do listy materiałów',
+  },
+};
+
+export const MaterialsCard = ({ status = 'idle', ...props }) => {
+  return (
+    <Box status={status} w="100%" boxShadow="base" borderRadius="8px" padding="24px" {...props}>
```
KonradSzwarc, line 32: 🤔 I'm not sure the Box component can receive a `status` prop. Why did you need to pass it here?

packages/ui/src/components/MaterialsCard/MaterialsCard.tsx
```diff
@@ -0,0 +1,53 @@
+import React from 'react';
+import { Box, BoxProps } from '@chakra-ui/react';
+
+import { Button } from '../Button';
+import { Typography } from '../Typography/Typography';
```
KonradSzwarc, line 5: 🙃 `import { Typography } from '../Typography';`

KonradSzwarc (changes requested):

talcia:
> @talcia When creating a PR remember to add "Closes #ISSUE_ID" to its description

Manually linking a pull request to an issue isn't enough?

packages/ui/src/components/MaterialsCard/MaterialsCard.test.tsx
```diff
@@ -0,0 +1,31 @@
+import React from 'react';
+import { render, screen } from '@testing-library/react';
+
+import { MaterialsCard } from './MaterialsCard';
+
+describe('MaterialsCard', () => {
+  it('should renders correctly in idle status', () => {
+    render(<MaterialsCard status="idle" />);
+
+    const button = screen.getByText('Wygeneruj listę');
```
hoolek77, line 10: 🤔  Shouldn't we import those texts? We've already defined them in the component file

hoolek77 (changes requested):

packages/ui/src/components/MaterialsCard/MaterialsCard.tsx
```diff
@@ -0,0 +1,53 @@
...
+
+type MaterialsCardStatus = 'idle' | 'loading' | 'generated';
```
mrpodkalicki, line 7: 💭 I think better way is used ,,const enum'' because we used this value in a few spots (MaterialsCard, MateriaslCard.test, MaterialCard.stories) so if we'll want to change somtehing in future it will be easier. 
Second reason  is we don't have to  create again status object in MateriaCard.stories.tsx we'll just assign const enum. (line 7)

const MaterialsCardStatus = {
  idle: 'idle',
  loading: 'loading',
  generated: 'generated',
} as const;

hoolek77 (approved):

KonradSzwarc (approved):

KonradSzwarc:
> @KonradSzwarc Manually linking a pull request to an issue isn't enough?

@talcia This way you link it only in ZenHub. When you write "Closes #ISSUE_ID" it's linked both in ZenHub task and GitHub issue. You can see the difference from the issue perspective below.
CleanShot 2021-07-28 at 16 36 51@2x

packages/ui/src/components/MaterialsCard/MaterialsCard.tsx
KonradSzwarc: I'm not sure if it's beneficial as in such a case we would need to always have two imports (component and enum). The question is, how it makes future changes easier? If we change the name of one value we will need to replace it in every place it's used anyway.

mrpodkalicki: yes, but in my opinion, it is better to change the variable than a string and with this convention never miss a place that needs to be updated

KonradSzwarc: It seems like a really good argument when you create a JavaScript app where misspelling some string or forgetting to update it somewhere are quite common problems. I think when it comes to TypeScript those problems cannot happen as you will be notified in the console (and probably in your IDE as well).

PR #117, opened 2021-08-01
feat(api): create complete user authentication flow (#116)

### LINKED ISSUE #116: Create complete user authentication flow

KonradSzwarc:
Closes #116


packages/api/src/auth/auth.controller.spec.ts
```diff
@@ -0,0 +1,42 @@
...
+  beforeEach(async () => {
+    const module = await Test.createTestingModule({
+      controllers: [AuthController],
+    }).compile();
+
+    controller = await module.resolve(AuthController);
+  });
+
+  describe('logout', () => {
+    it('Returns true for successful requests', async () => {
```
nowakprojects, line 21: I don't know why we need response body here. true/ false do not have some exact meaning. 
I don't see any test for not successful request.
For me response codes are enough for indicating that request was successful (200), or not (400 for example).

packages/api/src/auth/github/github.service.spec.ts
```diff
@@ -0,0 +1,65 @@
...
+  email: 'example@test.com',
+  image: 'https://photo-url.com',
+});
+
+describe('Github controller', () => {
+  let service: GithubService;
+  let usersRepository: Partial<UsersRepository>;
+
+  beforeEach(async () => {
+    usersRepository = {
```
nowakprojects, line 23: What makes me worry a lot it's that here we use users module internals (UsersRepository). 
If we want to achieve true modularization, we should treat each module as a separate service with a designed API to communicate with. Or maybe if something cannot work without something else - we should explicit make one module from that.

In this case, modules do not make sense if you can in module GitHub use whatever you want (repository - which is an implementation detail of module Users).
Moreover, you have here circular dependency (logically) in modules: you use UsersRepository in GitHub and githubUserId in UsersRepository.

The quickest solution, which comes to my mind, is: 
- publish event when new user registered / created - UserWasRegistered / UserWasCreated - which is public API of users module. With NestJS using events is so easy: https://docs.nestjs.com/recipes/cqrs (full CQRS architecture not required for that).
- listen to this event and in GitHubModule keep local copy of all users (but only what is required in GitHub Module) - and use that copy to use cases around GitHub.
This give you that those modules will not be tightly coupled. UsersModule may know nothing about GitHubModule and GitHubModule know about UsersModule only the event UserWasRegistered. After that you can refactor etc. (change persistence, how do you save users, users model fields / behaviors) without worrying about that different module (GitHub in this example) using something from it. You need just to keep contract which is an event.

With this approach, you have data duplication, of course, but it's not a problem. Think of that as microservices- in microservices you also keep local copies in order to not call another modules which may cause disaster if one of them is unavailable. In monolith distaster is in different cause - without clean boundaries you will come to moment when everything use everything - all modules are tightly couples and in fact there is one big module - Big Ball of Mud with low cohesion between connected modules / classes. Monolith is anti-pattern nowadays and we should go in direction of Modular-Monolith. 

Nice article: http://www.kamilgrzybek.com/design/modular-monolith-primer/


I'm not know if it's clear? Let me know if you have questions regarding these comments. But the most important rule is to keep context and modularization in the project. Without that, we can only make CRUD for database, but not model real business processes without creating spaghetti code without unmaintable dependencies. In this case, the mistake is not huge, but it's important to start with some best practices as soon as possible and treat them as convention.

You will gain another benefit - with clean APIs and not using another module's internals, you can easily extract one of the modules as microservice with almost no refactoring (you just change that events do not use in process eventbus, but something like Kafka / RabbitMQ).

nowakprojects (dismissed):
We need to define what it's "module" and what we want to achieve. More in the comment for github.service.spec.ts.

packages/api/src/auth/auth.controller.spec.ts
KonradSzwarc: You're right. I will remove the return and send 204 to specify that it's a successful response but with no content.

When it comes to testing non-successful requests I don't think there is a need to do so. This controller only passes errors created by services it's using - that's the default nest behavior.  In my opinion, we should test errors only in places where we throw them.

packages/api/src/auth/github/github.service.spec.ts
KonradSzwarc: Thank you for such a detailed description 🙂 It's hard to disagree that using patterns you described will result in more maintainable code - especially for a large app. Saying so, I think you missed some factors that exist only in the context of our team:

1. Experience - most of our team members are just starting their backend journey. Implementing CQRS in such a case will result in many additional hours of learning on their side, longer code reviews, and maybe even situations where someone decides to leave the backend team (we already had such situations in the frontend team). Also, such advanced concepts will create much higher entry barriers for new developers.
2. Deadline - CodersCamp starts in a few months, and that's our time to create the entire app. As a result, we need to search for compromises between code scalability and the speed of its development.
3. Scale - first version of an app will be rather small, so I don't think we can encounter any problems related to the current architecture. As of now, we want to have everything tested, I think we can refactor its structure when CodersCamp starts.

If you think there are ways to overcome those obstacles (especially the first one), you can suggest them. I think the entire team will be more than grateful if we can learn more best practices right away.

P.S. It might be a great place to nudge @pietruszka - you worked with CQRS in Mali Wspaniali, so maybe you have some thoughts about implementing it in a team of developers with various levels of experience.

KonradSzwarc: When it comes to this particular use case I also think there is nothing strange that authorization and users are connected modules. I agree it's maybe not the best solution but for sure (as tutorials and NestJS docs taught me 😅) it's the default one.

nowakprojects: It has nothing TODO with CQRS, just NestJS named in this way a module with have nice support for in-process events.

Ad.1, 2
I agree the greatest challenge may be the team members' experience. But... 
I hope that its possible to learn finding appropriate modules boundaries without doing it wrong for severals years (like in my career) by thinking in terms of database tables / documents. 

If someone can understand:
- how to do modular monolit (how to find modules boundaries - as small as possible). 
- Basic of Ports & Adapters Architecture 
- database design should not drive code design
- usage of events. 



Its enoguh to grasp more advanced concepts later. So, maybe short workshop next week? And everyone can try if its possible. 


No CQRS, no Event Sourcing etc. At all. 
Ad. 3 Test will be easier and more stable when implementation changes, becuase now it will be difficult to refactor, in test we need to know GitHub dependencies and even mock them. Such approach should speed up whole Development procesess and result in less bugs (I believe its possible regarding to 1. I've done it with my team (most of them, more backend focused) on CodersCamp). 

Go on with that feature - its just authentication and not whole application nerd equal Architecture. Most important are core functionalities.

nowakprojects: Some small, real feature for the app, like generating process st list would be perfect /or anything else with smilar scope.

kacper-cyra (approved):
Everything looks fine. I'm going only to mention one thing that we missed, Github user by default have his email access set to private and also don't have a name.

kacper-cyra (approved):

mecies (approved):
looks good to me ✅

PR #118, opened 2021-08-01
feat(website): Create Navbar section (#90)

### LINKED ISSUE #90: Create Navbar section
**Mockups:**
[Navbar](https://www.figma.com/file/ur0KLA9ZOtiodrAmpXrxh9/CodersCamp-app-and-website?node-id=801%3A1936)
[Navbar Item](https://www.figma.com/file/ur0KLA9ZOtiodrAmpXrxh9/CodersCamp-app-and-website?node-id=875%3A2840)

KrystianKjjk:
- Created navbar 
- Added empty pages for mentor, FAQ, contact 

closes #90


KonradSzwarc (changes requested):

packages/ui/package.json
```diff
@@ -16,6 +16,7 @@
     "react": "17.0.2"
   },
   "dependencies": {
+    "@chakra-ui/icons": "1.0.14",
```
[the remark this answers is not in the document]
KrystianKjjk: I needed sth for "hamburgerMenu" icon and if we use chakra anyway, so it won't be a problem I thought

packages/website/src/components/Navbar/Navbar.tsx
```diff
@@ -0,0 +1,26 @@
...
+import { useBreakpointValue } from '@coderscamp/ui/hooks/useBreakpointValue';
+
+import { DesktopBaseNavbar, MobileBaseNavbar } from './BaseNavbar';
+
+export const Navbar = () => {
+  const baseNavbar = useBreakpointValue({ base: <MobileBaseNavbar />, lg: <DesktopBaseNavbar /> } as const);
+
+  return (
+    <HStack
+      padding="18px calc(4px + 2.2vw)"
```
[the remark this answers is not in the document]
KrystianKjjk: Im not sure, but ok, simple 2.4vw looks good too I think

packages/website/src/components/Navbar/BaseNavbar.tsx
```diff
@@ -0,0 +1,67 @@
...
+
+export const MobileBaseNavbar = () => {
+  const logoLayout = useBreakpointValue({ sm: 'square', md: 'horizontal' } as const);
+  const { route } = useRouter();
+  const fontWeight = (link: string) => (link === route ? 'bold' : 'normal');
+
+  return (
+    <Flex>
+      <ChakraMenu>
+        <MenuButton as={IconButton} aria-label="Options" icon={<HamburgerIcon />} />
```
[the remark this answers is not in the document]
KrystianKjjk: I see it is centered and on the same level as buttons on the right side 🤔

packages/website/src/components/Navbar/BaseNavbar.tsx
```diff
@@ -0,0 +1,67 @@
...
+import { NavbarItem } from './NavbarItem';
+
+const NavbarElements = [
+  {
+    text: 'Strona główna',
+    destinationLink: '/',
+  },
+  {
+    text: 'Dla mentorów',
+    destinationLink: '/forMentors',
```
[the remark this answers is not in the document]
KrystianKjjk: Seems like that, but routing is based on file names, so if I call route "mentorzy" we will have file name "mentorzy.tsx" 
or there is a way to modify that default behavior?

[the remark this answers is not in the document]
KonradSzwarc: Our website is only one language so it wouldn't be a problem if you change filenames to Polish

packages/website/src/components/Navbar/BaseNavbar.tsx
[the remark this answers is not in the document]
KonradSzwarc: This is how it looks on my mobile view. Maybe this problem occurs only on a certain viewport width
CleanShot 2021-08-02 at 20 36 46@2x

packages/ui/package.json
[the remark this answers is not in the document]
KonradSzwarc: We use our own set of icons instead of Chakra ones. My bad because I didn't add the burger icon to Figma. Now you can find it [here](https://www.figma.com/file/ur0KLA9ZOtiodrAmpXrxh9/CodersCamp-app-and-website?node-id=1082%3A3353). Copy both solid and outlined versions. As they are the same you can use any of them 😄

packages/website/src/components/Navbar/BaseNavbar.tsx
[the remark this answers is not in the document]
KrystianKjjk: Ok, it happened in one case, when "CC" appears

How did u notice it...

packages/website/src/components/Navbar/BaseNavbar.tsx
```diff
@@ -0,0 +1,67 @@
...
+            </Link>
+          ))}
+        </MenuList>
+      </ChakraMenu>
+      {logoLayout && <Logo ml="6" maxWidth="280px" color="black" layout={logoLayout} />}
+    </Flex>
+  );
+};
+
+export const DesktopBaseNavbar = () => {
```
mrpodkalicki, line 54: 🤔 Maybe we should consider add boundary for navbar,  when screen is extra big.

KonradSzwarc (dismissed):

hoolek77 (approved):

packages/website/src/components/Navbar/Navbar.tsx
```diff
@@ -0,0 +1,25 @@
...
+import { useBreakpointValue } from '@coderscamp/ui/hooks/useBreakpointValue';
+
+import { DesktopBaseNavbar, MobileBaseNavbar } from './BaseNavbar';
+
+export const Navbar = () => {
+  const baseNavbar = useBreakpointValue({ base: <MobileBaseNavbar />, lg: <DesktopBaseNavbar /> } as const);
+  const buttonSize = useBreakpointValue({ base: 'xs', sm: 'md' } as const);
+
+  return (
+    <HStack padding="18px 2.4vw" width="100%" justifyContent="space-between" alignItems="center" position="relative">
```
KonradSzwarc, line 13: 💭 `18px 2.4vw` still seems a little tricky. It would be great to use pixels of the vertical padding.

packages/website/src/components/Navbar/Navbar.tsx
```diff
@@ -0,0 +1,25 @@
...
+      {baseNavbar}
+      <Flex>
+        <Button size={buttonSize} mr="3">
```
KonradSzwarc, line 16: 💭 To be more consistent you can use value with the `px` suffix for the margin-right property

packages/website/src/components/Navbar/Navbar.tsx
KrystianKjjk: If i use pixels for the vertical padding it will be too big for mobile or too small on desktop. I can use UseBreakpointValue() but i wouldn't say that additional logic will be better than simple vw

KonradSzwarc: Pixel values for spacing can be defined on a design system level and then unified across the entire app. When you use magic values like 2.4vw it makes space sizing unpredictable and less maintainable (imagine many components with such type of dynamic padding contained in one layout).

Of course, I agree that Chakra's way of making breakpoints creates an additional logic layer in the code. However, it's still easier to test layouts that may potentially beak only on five predefined breakpoints than layouts with many dynamic-sized elements that can break on any size of the screen.

KonradSzwarc: Just to clarify. I'm not saying that dynamic sizing is always bad. In cases where you want to divide the layout into many parts and for example use `width: 25vw` it's totally ok. In my opinion, the problem emerges when you create some dynamic values only because a particular, very precise number seems to work in a certain context.

KrystianKjjk: I understand, still not convinced that using Chakra's hook is better in this case, but i changed to pixels

KonradSzwarc (approved):

hoolek77 (approved):

KonradSzwarc (approved):

kacper-cyra (approved):

packages/website/src/components/Navbar/BaseNavbar.tsx
mrpodkalicki: Navbar is set on the full width of a page, for high resolution f.e 2500px, it doesn't look good because logo and buttons are near to the edge of the page. The rest of the content is centre and doesn't take full with of page. Maybe we should add a boundary f.e above 2200px resolution navbar should be not on the full width.

navbar

KrystianKjjk: Good point, corrected

packages/ui/icons/SolidMenu.svg
```diff
@@ -1,3 +1,3 @@
 <svg width="24" height="24" viewBox="0 0 24 24" fill="none" xmlns="http://www.w3.org/2000/svg">
 <path d="M3 4H21V6H3V4ZM3 11H21V13H3V11ZM3 18H21V20H3V18Z" fill="black"/>
-</svg>
+</svg>
```
KonradSzwarc, line 3: 🙃 GitHub doesn't like when there is no newline

packages/website/src/components/Navbar/Navbar.tsx
```diff
@@ -0,0 +1,36 @@
...
+import { Flex } from '@coderscamp/ui/components/Flex';
+import { HStack } from '@coderscamp/ui/components/Stack';
+import { useBreakpointValue } from '@coderscamp/ui/hooks/useBreakpointValue';
+
+import { DesktopBaseNavbar, MobileBaseNavbar } from './BaseNavbar';
+
+export const Navbar = () => {
+  const baseNavbar = useBreakpointValue({ base: <MobileBaseNavbar />, lg: <DesktopBaseNavbar /> } as const);
+  const buttonSize = useBreakpointValue({ base: 'xs', sm: 'md' } as const);
+  const horizontalPadding = useBreakpointValue({ base: '8px', sm: '18px', md: '24px', lg: '40px' });
```
KonradSzwarc, line 12: 🙃 Padding supports responsive values so there is no need for `useBreakpointValue`

KonradSzwarc (approved):

packages/website/src/components/Navbar/BaseNavbar.tsx
mrpodkalicki: 🙃 It'll be great if a navbar has been the same width as the title, but it's just a cosmetic. I give approval for it.

mrpodkalicki (approved):

KonradSzwarc (approved):

kamildzierzak (approved):

PR #120, opened 2021-08-03
feat(api): create endpoint for the initial form (#67)

### LINKED ISSUE #67: Create endpoint for the initial form
**Acceptance criteria:**
- There is an `POST /users/initial-form` endpoint.
- The endpoint takes data structure defined in cooperation with the author of #3.
- Basic data is saved as a part of the User entity. It contains full name, email, city, and sex.
- Rest of the data is saved as a separate entity and connected to the user with one-to-one relation.
- All related data models are placed in the `shared` directory.

kacper-cyra:
closes #67

### DIFF
--- a/packages/api/src/users/queries/handlers/find-user-with-given-github-id.handler.ts
+++ b/packages/api/src/users/queries/handlers/find-user-with-given-github-id.handler.ts
@@ -3,7 +3,7 @@ import { IQueryHandler, QueryHandler } from '@nestjs/cqrs';
 import { RegisteredUser, User } from '@coderscamp/shared/models';
 
 import { UsersRepository } from '../../users.repository';
-import { FindUserWithGivenGithubIdQuery } from '..';
+import { FindUserWithGivenGithubIdQuery } from '../find-user-with-given-github-id.query';
 
 @QueryHandler(FindUserWithGivenGithubIdQuery)
 export class FindUserWithGivenGithubIdQueryHandler implements IQueryHandler<FindUserWithGivenGithubIdQuery> {


packages/api/src/contracts/user.repository.service.ts
```diff
@@ -0,0 +1,15 @@
+import { RegisteredUser, Survey, User, UserInformation } from '@coderscamp/shared/models/user';
+
+export abstract class UserRepositoryService {
```
nowakprojects, line 3: Nice abstraction over Prisma - good :) 
But, This name... "UserRepository" is enough. Service / Repository. 

More: Repository should be per entity. For user or for survey - no for both. 
You don't need relation. If everything is related to user and doing by mean of user, then you will end up with methods only about user. 

I necourage you to read first email from: 
https://zycienakodach.pl/lista-mailingowa

packages/api/src/prisma/prisma.user.adapter.ts
```diff
@@ -0,0 +1,43 @@
...
+
+  async saveSurvey({ id, Survey: survey, ...user }: SharedUser) {
+    return this.prisma.user.update({
+      where: { id },
+      data: {
+        ...user,
+        Survey: { create: { ...survey } },
+      },
+      include: { Survey: true },
+    }) as unknown as SharedUser;
```
KonradSzwarc, line 37: 🤔 Looks like a problem. Why do you need to do type casting here?

packages/api/src/users/users.controller.spec.ts
```diff
@@ -0,0 +1,100 @@
...
+
+    await Promise.all([app.init(), db.migrate()]);
+    await repository.create({
+      githubId: 12345678,
+      fullName: 'Name',
+      email: 'example@test.com',
+      image: 'https://photo-url.com',
+    });
+
+    userSurvey = {
```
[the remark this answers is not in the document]
nowakprojects: I recommend get rid of survey references from users module at all.
We need two-way relation. Leave just reference by userId from Survey to User. 
There is no business rule between user data like email / gender and the survey, so we can split it.

If we go this way, we will end up references to all other modules in users and users test, because almost everything in the system is related to some user. The test will blow up with things totally unrelated to users and any changes / new features requires changes in users module with is totally oposite of Open-Closed.

packages/api/src/users/users.repository.ts
```diff
@@ -1,22 +1,35 @@
...
 
-import type { User } from '@coderscamp/shared/models/user';
+import type { RegisteredUser, User, UserInformation } from '@coderscamp/shared/models/user';
 
-import { PrismaService } from '../prisma/prisma.service';
+import type { GithubUserData } from '../auth/github/github.model';
+import { UserRepositoryService } from '../contracts/user.repository.service';
 
 @Injectable()
 export class UsersRepository {
```
nowakprojects, line 9: I don't undesrtand why we have this class. Why do not use 'UserRepositoryService' (renamed it to UsersRepository) directly?

packages/api/src/users/commands/handlers/save-survey.handler.ts
```diff
@@ -0,0 +1,23 @@
...
+import { SaveSurveyCommand } from '../save-survey.command';
+
+@CommandHandler(SaveSurveyCommand)
+export class SaveSurveyHandler implements ICommandHandler<SaveSurveyCommand> {
+  constructor(private repository: UsersRepository) {}
+
+  async execute({ input }: SaveSurveyCommand): Promise<User> {
+    const userFromDb = await this.repository.getUser(input.id);
+
+    if (!userFromDb) throw new NotFoundException(ErrorMessage.USER_NOT_FOUND);
```
nowakprojects, line 17: I think is not an issue to have survey for user which not exist.
More problematic is that we need to load a whole user entity just to save survey. 
If we follow this approach in the future, we will end up with loading all users related things (projects / comments) etc. in every case (lazy loading is even more evil - difficult to keep invariants / business rules).

I'd recommend removing check is user exist. 
And the next one about SURVEY_ALREADY_FILLED hmmm...
There is a short period of time between line 15 (load)  and 19, so while loading from db survey may not be filled, but in 19 yes, and the code will not detect that.  
We should decide what to do: 
- use optimistic locking
- override the survey

packages/api/src/users/users.controller.ts
```diff
@@ -1,16 +1,22 @@
...
+
+  @Post('survey')
+  @UsePipes(new YupValidationPipe(userSchema))
+  async saveUserSurvey(@Body() userSurveyDTO: UserSurveyDTO): Promise<UserSurveyDTO> {
+    const command = new SaveSurveyCommand(UsersMapper.userSurveyToDomain(userSurveyDTO));
 
-  @Get('/')
-  async getAll(): Promise<UserDTO[]> {
-    return UsersMapper.toPlainMany(await this.usersRepository.getAll());
+    return UsersMapper.userSurveyToPlain(await this.commandBus.execute(command));
```
nowakprojects, line 20: I don't see any value in returning the whole object.

kacper-cyra:
I have some questions about what should be done.
@KonradSzwarc  sent link on Slack to GitHub repository with testing best practices with recommends using database in docker as it's usually the most similar solution to actual database. In our case I already went for in memory database which is faster than docker but it introduces more problems: migrations, SQL queries. So I wanted to ask about yours opinion on this topic.

nowakprojects:
> I have some questions about what should be done.
> @KonradSzwarc sent link on Slack to GitHub repository with testing best practices with recommends using database in docker as it's usually the most similar solution to actual database. In our case I already went for in memory database which is faster than docker but it introduces more problems: migrations, SQL queries. So I wanted to ask about yours opinion on this topic.

Yes, if we need a lot of additional work like SQL, is not worth it. While talking about (in-memory i was thinking about something else, but I've learned also was is possible from this PR :D )
Another approach, faster / easier is that: use docker only for integration tests, but for unit tests (where repository is needed) use in-memory implementation, which I mean implementation which is in-process. Like implementation on object / list: https://github.com/nowakprojects/CodersCamp2020.Project.FullStack-Node-React.TableSoccerTournaments/blob/develop/backend/src/modules/doubles-tournament/infrastructure/repository/inmemory/InMemoryDoublesTournamentRepository.ts 

Simple, fast. No need for db cleanup before each unit test. 
It's TS version of what Kuba Nabrdalik described here: https://youtu.be/ILBX9fa9aJo?t=3115

packages/api/src/auth/github/github.mapper.ts
```diff
@@ -0,0 +1,15 @@
+import { RegisteredUser } from '@coderscamp/shared/models/user';
+
+import { GithubUser } from './github.types';
+
+export class GithubMapper {
+  static fromGithubToDomain(user: GithubUser): Omit<RegisteredUser, 'id'> {
```
nowakprojects, line 6: Is this valid object? Can we have "RegisteredUser" without it? If id is not set is the user already "registered"? 
Looks like we have something which can be assured by types (like every user has an id), but is not here.

packages/api/src/auth/github/github.repository.ts
```diff
@@ -0,0 +1,22 @@
...
+
+  async findUserById(id: number) {
+    return this.repository.findById(id);
+  }
+
+  async findUserByGithubId(id: number) {
+    return this.repository.findByGithubId(id);
+  }
+
+  async createUser(data: Omit<RegisteredUser, 'id'>) {
```
nowakprojects, line 19: I recommend use of something like UUID as ID and get rid of handling situation when id is omitted. 
It's also better for app scalability than leaving generating id for the database.

packages/api/src/auth/github/github.service.ts
```diff
@@ -1,19 +1,23 @@
...
+import { GithubRepository } from './github.repository';
 
 @Injectable()
 export class GithubService {
-  constructor(private readonly usersRepository: UsersRepository) {}
+  constructor(private readonly githubRepository: GithubRepository) {}
 
-  async authorizeUser(githubUserData: UserFromGithub): Promise<User> {
-    let user = await this.usersRepository.getByGithubId(githubUserData.githubId);
+  async authorizeUser(githubUserData: Omit<RegisteredUser, 'id'>): Promise<RegisteredUser> {
```
nowakprojects, line 12: Once more: 
Omit<RegisteredUser, 'id'> 

looks like "Registered but not exactly, not already because do not have id" :P

packages/api/src/auth/jwt/jwt.mapper.ts
```diff
@@ -0,0 +1,15 @@
...
+export class JwtMapper {
+  static fromDomainToJwt(user: UserFromJwt): JwtPayload {
+    return {
+      sub: String(user.id),
+    };
+  }
+
+  static fromJwtToDomain(payload: JwtPayload): UserFromJwt {
+    return {
+      id: Number(payload.sub),
```
nowakprojects, line 12: It's better to use string for ids in domain. It leaves you flexibility if its a number / uuid or some natural key (like email).

packages/api/src/contracts/survey.repository.ts
```diff
@@ -0,0 +1,9 @@
+import type { Survey, UserSurvey } from '@coderscamp/shared/models';
+
+export abstract class SurveyRepositoryPort {
```
nowakprojects, line 3: Very good! But i think saveAndUpdateRelatedUser is not necessary. What is a use case for this method?

packages/api/src/contracts/user.repository.ts
```diff
@@ -0,0 +1,14 @@
...
+export abstract class UserRepositoryPort {
+  abstract create(data: Omit<RegisteredUser, 'id'>): Promise<User | RegisteredUser>;
+
+  abstract findByGithubId(githubId: number): Promise<User | RegisteredUser | null>;
+
+  abstract findById(id: number): Promise<User | RegisteredUser | null>;
+
+  abstract update(data: User): Promise<User | RegisteredUser>;
+
+  abstract getUser(userId: number): Promise<((UserSurvey | RegisteredUser) & { Survey: Survey | null }) | null>;
```
nowakprojects, line 13: I would get rid of UserSurvey at all from UserRepository. 
There is very low cohesion. UserSurvey is not required in every usecase, but will be fetched from database every time when you query for certain user.

packages/api/src/surveys/survey.repository.ts
```diff
@@ -0,0 +1,22 @@
+import { Injectable } from '@nestjs/common';
+
+import type { Survey, UserSurvey } from '@coderscamp/shared/models';
+
+import { SurveyRepositoryPort } from '../contracts/survey.repository';
+
+@Injectable()
+export class SurveyRepository {
```
nowakprojects, line 8: I'm not sure what is the purpose of this class? 
Why not make implementations of SurveyRepositoryPort @Injectable?

packages/api/src/contracts/github.repository.ts
```diff
@@ -0,0 +1,9 @@
+import { RegisteredUser, User } from '@coderscamp/shared/models/user';
+
+export abstract class GithubRepositoryPort {
```
nowakprojects, line 3: Why we have 'abstract class' here? 
What about interface?

kacper-cyra:
@nowakprojects  @KonradSzwarc  How do we want to handle data storing and loading for tests? I saw that Pirsma has its own test utils that uses faker.js to generate data on the fly and inserts it to the database.

packages/api/src/contracts/github.repository.ts
kacper-cyra: I don't think we can. Nest needs a class to replace. Unless we just explicitly say in each module which repository it should use. 

https://github.com/CodersCrew/coderscamp/blob/c8830ad87213d435a8feb428ba0402208ac94b0e/packages/api/src/prisma/prisma.module.ts#L11-L15

packages/api/src/contracts/survey.repository.ts
kacper-cyra: The idea is just save the initial form in one go. There is no special reason for doing so this way as from what I read Prisma makes 2 separate queries anyway.

packages/api/src/contracts/github.repository.ts
nowakprojects: class is not required, you can use string or symbol (i prefer the second one): 

https://docs.nestjs.com/fundamentals/custom-providers

Example you can find here: 
https://github.com/CodersCrew/coders-board-v2/blob/master/backend/src/inviting-applicants/write-side/infrastructure/inviting-applicants-infrastructure.module.ts

with symbol APPLICANT_INVITATION_REPOSITORY

packages/api/src/contracts/survey.repository.ts
nowakprojects: So if 2 separate queries - no optimization here, it's better to do not have coupling between Survey and User just for... what :) ? We don't have any advantages, but we have blurred survey module with user concept.

PR #121, opened 2021-08-04
feat(ui): create timeline item component (#111)

### LINKED ISSUE #111: Create TimelineItem component
**Mockups:**
[Figma link](https://www.figma.com/file/ur0KLA9ZOtiodrAmpXrxh9/CodersCamp-app-and-website?node-id=784%3A3438)

KrystianKjjk:
closes #111

### DIFF
--- a/packages/ui/src/components/TimelineItem/TimelineItem.stories.tsx
+++ b/packages/ui/src/components/TimelineItem/TimelineItem.stories.tsx
@@ -12,7 +12,7 @@ const meta: Meta = {
 export default meta;
 
 const Template: Story<TimelineItemProps> = (args) => (
-  <Box bg="gray.200" p="40px">
+  <Box bg="white" p="40px">
     <TimelineItem {...args} width="400px" height="292px" mb="10px" />
   </Box>
 );
--- a/packages/ui/src/components/TimelineItem/TimelineItem.stories.tsx
+++ b/packages/ui/src/components/TimelineItem/TimelineItem.stories.tsx
@@ -1,0 +1,34 @@
+import React from 'react';
+import { Meta, Story } from '@storybook/react';
+
+import { TimelineItem, TimelineItemProps } from './TimelineItem';
+
+const tags = ['h1', 'h2', 'h3', 'h4', 'h5', 'h6', 'a', 'span', 'p', 'div'];
+const sizes = ['xs', 'sm', 'md', 'lg', 'xl', '2xl', '3xl', '4xl', '5xl', '6xl', '7xl', '8xl'];
+
+const meta: Meta = {
+  title: 'Components / TimelineItem',
+  component: TimelineItem,
+  argTypes: {
+    as: {
+      control: { type: 'select' },
+      options: tags,
+    },
+    size: {
+      control: { type: 'select' },
+      options: sizes,
+    },
+  },
+};
+
+export default meta;
+
+const Template: Story<TimelineItemProps> = (args) => <TimelineItem {...args} />;
+
+export const Playground = Template.bind({});
+
+Playground.args = {
+  step: 0,
+  date: '24.10-11.11.2021',
+  information: 'lorem ipsum ',
+};
--- a/packages/ui/src/components/TimelineItem/TimelineItem.stories.tsx
+++ b/packages/ui/src/components/TimelineItem/TimelineItem.stories.tsx
@@ -1,34 +1,27 @@
 import React from 'react';
+import { Box } from '@chakra-ui/react';
 import { Meta, Story } from '@storybook/react';
 
 import { TimelineItem, TimelineItemProps } from './TimelineItem';
 
-const tags = ['h1', 'h2', 'h3', 'h4', 'h5', 'h6', 'a', 'span', 'p', 'div'];
-const sizes = ['xs', 'sm', 'md', 'lg', 'xl', '2xl', '3xl', '4xl', '5xl', '6xl', '7xl', '8xl'];
-
 const meta: Meta = {
   title: 'Components / TimelineItem',
   component: TimelineItem,
-  argTypes: {
-    as: {
-      control: { type: 'select' },
-      options: tags,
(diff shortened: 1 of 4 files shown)


packages/ui/src/components/TimelineItem/TimelineItem.tsx
```diff
@@ -0,0 +1,56 @@
...
+import { Typography } from '../Typography';
+
+export interface TimelineItemProps extends FlexProps {
+  step: number;
+  date: string;
+  information: string;
+}
+
+export const TimelineItem: React.FC<TimelineItemProps> = ({ step, date, information, ...props }) => {
+  const boxPadding = useBreakpointValue({ base: '18px', md: '32px' } as const) ?? '18px';
```
KonradSzwarc, line 13: 💭 You can define the default breakpoint as the second argument of the `useBreakpointValue`.

packages/ui/src/components/TimelineItem/TimelineItem.tsx
```diff
@@ -0,0 +1,56 @@
...
+        fontWeight={800}
+      >
+        {step}
+      </Center>
+      <Typography
+        my={dateVerticalMargin}
+        fontFamily="inter"
+        fontWeight={800}
+        fontSize="20px"
+        letterSpacing="-0.017em"
```
KonradSzwarc, line 45: 🤔 There shouldn't be the need to specify those properties separately. As you can see on Figma it's xl size with extrabold weight

packages/ui/src/components/TimelineItem/TimelineItem.tsx
```diff
@@ -0,0 +1,56 @@
...
+        color="gray.900"
+      >
+        {date}
+      </Typography>
+
+      <Typography fontFamily="inter" letterSpacing="0.014em" color="gray.700">
```
KonradSzwarc, line 51: 🤔 Same as in the comment above. It's just lg size typography

packages/ui/src/components/TimelineItem/TimelineItem.tsx
```diff
@@ -0,0 +1,56 @@
...
+    <Center
+      bg="white"
+      borderRadius="8px"
+      p={boxPadding}
+      pb="16px"
+      minHeight="248px"
+      flexDirection="column"
+      justifyContent="flex-start"
+      textAlign="center"
+      overflow="auto"
```
KonradSzwarc, line 26: 🤔 On mockups I see that this element has a border

packages/ui/src/components/TimelineItem/TimelineItem.test.tsx
```diff
@@ -0,0 +1,43 @@
+import React from 'react';
+import { render, screen } from '@testing-library/react';
+
+import { TimelineItem } from './TimelineItem';
+
+describe('TimelineItem', () => {
+  Object.defineProperty(window, 'matchMedia', {
```
KonradSzwarc, line 7: ❓Would it break the test if you don't define it?

packages/ui/src/components/TimelineItem/TimelineItem.stories.tsx
```diff
@@ -0,0 +1,27 @@
...
+
+const meta: Meta = {
+  title: 'Components / TimelineItem',
+  component: TimelineItem,
+};
+
+export default meta;
+
+const Template: Story<TimelineItemProps> = (args) => (
+  <Box bg="gray.200" p="40px">
```
KonradSzwarc, line 15: 💭 On the website it would be displayed on a white background so I think you can use the same bg color here

KonradSzwarc (changes requested):

packages/ui/src/components/TimelineItem/TimelineItem.test.tsx
KrystianKjjk: Yes, it's necessary always when you use useBreakpointValue() in component. useBreakpoint() uses this props inside, thats why whenever we use useBreakPoint() and useBreakpointValue() we have to mock this, i can move Object.defineProperty to different function but... where ?

packages/ui/src/components/TimelineItem/TimelineItem.tsx
KrystianKjjk: I did it, but still returned 'const boxPadding: "18px" | "32px" | undefined'

packages/ui/src/components/TimelineItem/TimelineItem.tsx
```diff
@@ -0,0 +1,51 @@
...
+      borderColor="#D1D5DB"
```
celinawierzbicka, line 27: you can use `gray.300` value from theme

celinawierzbicka (approved):

packages/ui/src/components/TimelineItem/TimelineItem.tsx
```diff
@@ -10,8 +10,8 @@ export interface TimelineItemProps extends FlexProps {
 }
 
 export const TimelineItem: React.FC<TimelineItemProps> = ({ step, date, information, ...props }) => {
-  const boxPadding = useBreakpointValue({ base: '18px', md: '32px' } as const) ?? '18px';
-  const dateVerticalMargin = useBreakpointValue({ base: '16px', md: '24px' } as const) ?? '16px';
+  const boxPadding = useBreakpointValue({ base: '18px', md: '32px' } as const, '18px');
```
KonradSzwarc, line 13: ❓ Does the default value work? As I understand the intellisense description it should be `base` instead of `18px`
CleanShot 2021-08-04 at 22 53 44@2x

KrystianKjjk: Right, corrected

KonradSzwarc (approved):

hoolek77 (approved):

Veryyapeee (approved):

PR #126, opened 2021-08-06
feat(ui): create slider steps component (#112)

### LINKED ISSUE #112: Create SliderSteps component
**Mockups:**
[Slider Steps](https://www.figma.com/file/ur0KLA9ZOtiodrAmpXrxh9/CodersCamp-app-and-website?node-id=784%3A3303)
[Slider Step](https://www.figma.com/file/ur0KLA9ZOtiodrAmpXrxh9/CodersCamp-app-and-website?node-id=784%3A3208)

KrystianKjjk:
closes #112


packages/ui/src/components/SliderSteps/SliderStep.tsx
```diff
@@ -0,0 +1,25 @@
...
+export interface SliderStepProps {
+  current: boolean;
+}
+
+export const SliderStep = ({ current }: SliderStepProps) => {
+  const bgColor = current ? 'brand.500' : 'gray.300';
+  const horizontalMargin = current ? '2px' : '10px';
+
+  return (
+    <Box
```
KonradSzwarc, line 13: 💭 You can use the Circle component here

packages/ui/src/components/SliderSteps/SliderSteps.tsx
```diff
@@ -0,0 +1,38 @@
...
+  };
+
+  return (
+    <Center width="max-content">
+      <OutlinedArrowLeftIcon
+        {...arrowStyles}
+        onClick={() => onChangeIdx(selectedIdx !== 0 ? selectedIdx - 1 : lastIdx)}
+      />
+      {[...Array(count).keys()].map((idx) => (
+        <SliderStep key={idx} current={selectedIdx === idx} />
```
KonradSzwarc, line 30: 🤔 `onChange` should be called also when user clicks a particular dot

packages/ui/src/components/SliderSteps/SliderSteps.tsx
```diff
@@ -0,0 +1,38 @@
...
+      ))}
+      <OutlinedArrowRightIcon
```
KonradSzwarc, line 32: 🤔 Looking at mockups it should be an IconButton instance

packages/ui/src/components/SliderSteps/SliderStep.tsx
```diff
@@ -0,0 +1,25 @@
...
+      data-testid={current ? 'currentSliderStep' : 'baseSliderStep'}
+      borderRadius="50%"
+      backgroundColor={bgColor}
+      my="0px"
+      mx={horizontalMargin}
+      w="12px"
+      h="12px"
+      boxSizing="content-box"
+      {...(current && { borderColor: 'brand.200', borderWidth: '4px' })}
+    />
```
KonradSzwarc, line 23: 💭 As it's an interactive element it would be great to show a 'pointer' cursor when user hovers it (only for non-active steps)

packages/ui/src/components/SliderSteps/SliderStep.tsx
```diff
@@ -0,0 +1,25 @@
...
+  current: boolean;
+}
+
+export const SliderStep = ({ current }: SliderStepProps) => {
+  const bgColor = current ? 'brand.500' : 'gray.300';
+  const horizontalMargin = current ? '2px' : '10px';
+
+  return (
+    <Box
+      data-testid={current ? 'currentSliderStep' : 'baseSliderStep'}
```
KonradSzwarc, line 14: 💭 We can make the entire component more accessible and as a bonus, there would be no need for `data-testid`.
1. Add `role="tablist"` to the SliderSteps component.
2. Add `role="tab" tabindex="0"` here.
3. Add `aria-selected={current}` here.

KonradSzwarc (changes requested):

packages/ui/src/components/SliderSteps/SliderSteps.test.tsx
```diff
@@ -1,15 +1,79 @@
...
 describe('SliderSteps', () => {
+  const givenStepsCount = 5;
+  const lastDotIndex = givenStepsCount - 1;
+  const selectedIndex = 2;
+
+  const TestSliderComponent = ({ startSelectedIndex }: { startSelectedIndex: number }) => {
+    const [selected, setSelectedIndex] = useState(startSelectedIndex);
+
+    return <SliderSteps count={givenStepsCount} selectedIndex={selected} onChange={setSelectedIndex} />;
+  };
```
KonradSzwarc, line 15: 💭 It would be better to test only the component - not its implementation. In practical terms it means you render normal SliderSteps, mock the `onChange` function and check what argument was passed to it.

packages/ui/src/components/SliderSteps/SliderSteps.test.tsx
```diff
@@ -1,15 +1,79 @@
...
+
+  const getSelectedDotIndex = () =>
+    screen.getAllByRole('tab').findIndex((x) => x.getAttribute('aria-selected') === 'true');
+
+  const fireButtonClick = (buttonLabel: string) => {
+    const button = screen.getByLabelText(buttonLabel);
+
+    fireEvent.click(button);
```
KonradSzwarc, line 23: 💭 You can use the [user-event](https://github.com/testing-library/user-event) library. It will be a small improvement as it uses events that are more similar to the one's users cause.

packages/ui/src/components/SliderSteps/SliderSteps.tsx
```diff
@@ -2,36 +2,40 @@ import React from 'react';
...
-    h: '32px',
-    p: '6px',
-    borderRadius: '6px',
-    color: 'gray.800',
-  };
+  const goForward = () => onChange(selectedIndex !== lastIdx ? selectedIndex + 1 : 0);
+  const goBack = () => onChange(selectedIndex !== 0 ? selectedIndex - 1 : lastIdx);
+  const goToSelected = (index: number) => onChange(index);
+
+  if (selectedIndex < 0 || selectedIndex >= count)
```
KonradSzwarc, line 25: 🙃 Adding brackets to this if could be a little improvement when it comes to readability

packages/ui/src/components/SliderSteps/SliderSteps.tsx
```diff
@@ -2,36 +2,40 @@ import React from 'react';
...
-  const arrowStyles = {
-    w: '32px',
-    h: '32px',
-    p: '6px',
-    borderRadius: '6px',
-    color: 'gray.800',
-  };
+  const goForward = () => onChange(selectedIndex !== lastIdx ? selectedIndex + 1 : 0);
+  const goBack = () => onChange(selectedIndex !== 0 ? selectedIndex - 1 : lastIdx);
+  const goToSelected = (index: number) => onChange(index);
```
KonradSzwarc, line 23: 💭 You can check if the index is the current one and in such case don't call the `onChange` function

packages/ui/src/components/SliderSteps/SliderSteps.test.tsx
KrystianKjjk: Using this library is good idea in general, but in that case it will change only 
```diff
    fireEvent.click(button);
```
to
```diff
    userEvent.click(button);
```
(or maybe I don't know about sth?) 
and what's more, if I implement your suggestion about testing only the component - not its implementation, this function become useless and ready to remove

packages/ui/src/components/SliderSteps/SliderSteps.tsx
KonradSzwarc: My bad, I meant curly brackets:
```
if (selectedIndex < 0 || selectedIndex >= count) {
    throw new Error('SelectedIndex property should be positive number and less than count of dots');
}
```

packages/ui/src/components/SliderSteps/SliderSteps.test.tsx
KonradSzwarc: Why you couldn't test such a case? I think you can use `expect(onChange).not.toHaveBeenCalled()`

packages/ui/src/components/SliderSteps/SliderSteps.test.tsx
KonradSzwarc: It's more about a standard than the number of use cases. When someone will implement a similar test in the future he/she could look at your code and copy the solution. If we agree that using `userEvent` is a better approach than `fireEvent` we should start using it from the first suitable use case.

I don't understand why this function becomes useless when you apply my suggestion about testing `onChange`. You would still need to click something and check arguments of the called `onChange` function.

packages/ui/src/components/SliderSteps/SliderStep.tsx
```diff
@@ -0,0 +1,35 @@
+import React from 'react';
+import { Circle } from '@chakra-ui/react';
+
+export interface SliderStepProps {
+  current: boolean;
+  onClick: () => void;
+}
+const currentDotProps = {
+  backgroundColor: 'brand.500',
+  mx: '2px',
```
kamildzierzak, line 10: 🤔 Shouldn't it be 6px? I don't know if this was supposed to be the effect, but by clicking on the dots on the storybook, I noticed that the margins jump strangely when switching to active.

packages/ui/src/components/SliderSteps/SliderSteps.test.tsx
```diff
@@ -2,95 +2,122 @@ import React from 'react';
...
+
+const clickForwardArrow = () => {
+  userEvent.click(screen.getByLabelText('go forward button'));
+};
+
+const clickBackArrow = () => {
+  userEvent.click(screen.getByLabelText('go back button'));
+};
+
+const dots = () => screen.getAllByRole('tab');
```
KonradSzwarc, line 16: 🙃 It's a function so `getAllDots` would be a more suitable name

KonradSzwarc (approved):

KonradSzwarc (changes requested):
Failed CI

KonradSzwarc (approved):

kamildzierzak (approved):

KonradSzwarc (approved):

KonradSzwarc (approved):

kamildzierzak (approved):

PR #127, opened 2021-08-06
feat(website): create big numbers section (#93)

### LINKED ISSUE #93: Create BigNumbers section
**Mockups:**
[Figma link](https://www.figma.com/file/ur0KLA9ZOtiodrAmpXrxh9/CodersCamp-app-and-website?node-id=773%3A2480)

talcia:
closes #93

### DIFF
--- a/packages/website/src/components/BigNumbers/BigNumbers.tsx
+++ b/packages/website/src/components/BigNumbers/BigNumbers.tsx
@@ -5,22 +5,26 @@ import { VStack } from '@coderscamp/ui/components/Stack';
 
 import { StatsWithTitle } from './StatsWithTitle';
 
-const amounts = {
-  currentEdition: ['7.', '1516', '200', '6'],
-  previousEditions: ['6526', '1022', '372', '502'],
-};
+const currentEdition = [
+  { amount: '7.', label: 'edycja' },
+  { amount: '1 516', label: 'zadań i materiałów' },
+  { amount: '200', label: 'miejsc na kurs' },
+  { amount: '6', label: 'projektów do stworzenia' },
+];
 
-const labels = {
-  currentEdition: ['edycja', 'zadań i materiałów', 'miejsc na kurs', 'projektów do stworzenia'],
-  previousEditions: ['zgłoszeń', 'osób na kursie', 'wydanych certyfikatów', 'zakodowanych projektów'],
-};
+const previousEditions = [
+  { amount: '6 526', label: 'zgłoszeń' },
+  { amount: '1 022', label: 'osób na kursie' },
+  { amount: '372', label: 'wydanych certyfikatów' },
+  { amount: '502', label: 'zakodowanych projektów' },
+];
 
 export const BigNumbers = () => {
   return (
     <Flex justify="center" mx="auto" pt="40px" pb="80px" maxW="min(1120px, 100%)">
       <VStack spacing="80px" width="100%">
-        <StatsWithTitle amounts={amounts.currentEdition} labels={labels.currentEdition} title="BIEŻĄCA EDYCJA" />
-        <StatsWithTitle amounts={amounts.previousEditions} labels={labels.previousEditions} title="POPRZEDNIE EDYCJE" />
+        <StatsWithTitle statistics={currentEdition} title="BIEŻĄCA EDYCJA" />
+        <StatsWithTitle statistics={previousEditions} title="POPRZEDNIE EDYCJE" />
       </VStack>
     </Flex>
   );
--- a/packages/website/src/components/BigNumbers/BigNumbers.tsx
+++ b/packages/website/src/components/BigNumbers/BigNumbers.tsx
@@ -5,21 +5,22 @@ import { VStack } from '@coderscamp/ui/components/Stack';
 
 import { StatsWithTitle } from './StatsWithTitle';
 
-const amounts = [
-  ['7.', '1516', '200', '6'],
-  ['6526', '1022', '372', '502'],
-];
-const labels = [
-  ['edycja', 'zadań i materiałów', 'miejsc na kurs', 'projektów do stworzenia'],
-  ['zgłoszeń', 'osób na kursie', 'wydanych certyfikatów', 'zakodowanych projektów'],
-];
+const amounts = {
+  currentEdition: ['7.', '1516', '200', '6'],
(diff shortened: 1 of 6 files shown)


packages/website/src/components/BigNumbers/StatsWithTitle.tsx
```diff
@@ -0,0 +1,40 @@
...
+  const statisticFlexDirection = useBreakpointValue({ base: 'column', md: 'row' } as const);
+  const statisticSize = useBreakpointValue({ xl: 'min(1120px, 100%)', md: 'min(850px, 100%)' } as const);
+
+  const stackProps = { direction: statisticStackDirection, width: statisticSize };
+  const flexProps = { direction: statisticFlexDirection, width: statisticSize };
+
+  return (
+    <VStack spacing="32px" width="100%">
+      <Typography size="md" color="gray.400" weight="bold">
+        BIEŻĄCA EDYCJA
```
KrystianKjjk, line 26: On figma there's "Bieżąca edycja" for first Stats and "poprzednie edycje" for second one

packages/website/src/components/BigNumbers/BigNumbers.tsx
```diff
@@ -0,0 +1,26 @@
...
+import { StatsWithTitle } from './StatsWithTitle';
+
+const amounts = [
+  ['7.', '1516', '200', '6'],
+  ['6526', '1022', '372', '502'],
+];
+const labels = [
+  ['edycja', 'zadań i materiałów', 'miejsc na kurs', 'projektów do stworzenia'],
+  ['zgłoszeń', 'osób na kursie', 'wydanych certyfikatów', 'zakodowanych projektów'],
+];
```
KrystianKjjk, line 15: I think it would be better to pass amount and labels as an objects. Later in code you will use something like 'amount.currentEdition' instead of 'amount[0]' what is more readable, same suggestion with nested arrays

packages/website/src/components/BigNumbers/StatsWithTitle.tsx
```diff
@@ -0,0 +1,40 @@
...
+      <Stack {...stackProps}>
+        <Flex {...flexProps}>
+          <Statistic amount={amounts[0]} label={labels[0]} pb="32px" />
+          <Statistic amount={amounts[1]} label={labels[1]} pb="32px" />
+        </Flex>
+        <Flex {...flexProps}>
+          <Statistic amount={amounts[2]} label={labels[2]} pb="32px" />
+          <Statistic amount={amounts[3]} label={labels[3]} />
+        </Flex>
+      </Stack>
```
KonradSzwarc, line 37: 💭 I think it would be cleaner to use the Grid or SimpleGrid component re-exported from Chakra UI instead of those Stack, Flex, and `pb` passed to the Statistic component. Also in my opinion when statistics stacks vertical you can dive them a little larger vertical padding.

packages/website/src/components/BigNumbers/StatsWithTitle.tsx
```diff
@@ -0,0 +1,40 @@
...
+
+interface StatsWithTitleProps {
+  amounts: string[];
+  labels: string[];
+}
+
+export const StatsWithTitle = ({ amounts, labels }: StatsWithTitleProps) => {
+  const statisticStackDirection = useBreakpointValue({ base: 'column', xl: 'row' } as const);
+  const statisticFlexDirection = useBreakpointValue({ base: 'column', md: 'row' } as const);
+  const statisticSize = useBreakpointValue({ xl: 'min(1120px, 100%)', md: 'min(850px, 100%)' } as const);
```
KonradSzwarc, line 18: 🙃 For consistency with previous two functions you can write `md` first and `xl` second in order

KonradSzwarc (changes requested):

packages/website/src/components/BigNumbers/StatsWithTitle.tsx
```diff
@@ -0,0 +1,37 @@
...
+
+  return (
+    <VStack spacing="32px" width={statisticSize}>
+      <Typography size="md" color="gray.400" weight="bold">
+        {title}
+      </Typography>
+      <SimpleGrid
+        columns={{ xl: 4, lg: 4, md: 2, sm: 1 }}
+        spacingX={{ lg: '32px' }}
+        spacingY={{ base: '72px' }}
```
KonradSzwarc, line 27: ❓ As you have only one value here is the responsiveness object necessary?

packages/website/src/components/BigNumbers/BigNumbers.tsx
```diff
@@ -0,0 +1,27 @@
...
+
+const amounts = {
+  currentEdition: ['7.', '1516', '200', '6'],
+  previousEditions: ['6526', '1022', '372', '502'],
+};
+
+const labels = {
+  currentEdition: ['edycja', 'zadań i materiałów', 'miejsc na kurs', 'projektów do stworzenia'],
+  previousEditions: ['zgłoszeń', 'osób na kursie', 'wydanych certyfikatów', 'zakodowanych projektów'],
+};
```
KonradSzwarc, line 16: 💭 I think it could be more readable if you define statistics as two arrays of objects (e.g. `currentEdition` and `previousEditions`). Each object in the array could use `StatisticProps` interface. Doing it this way gives you the following benefits:
1. In `StatsWithTitle.tsx` you can simply map the array instead of creating four Statistic elements manually.
2. StatsWithTitle can receive only two props - `title` and `statistics`.
3. You can destructure objects when passing them to the Statistic component instead of defining each prop separately.

Veryyapeee:
💭 I wonder if we could animate the numbers, something like, when we scroll to the section the animations goes like fast counter from 0 to destination number, what do you think @KonradSzwarc ?
Here's simple example of what I was thinking about https://css-tricks.com/animating-number-counters/

KonradSzwarc:
> 💭 I wonder if we could animate the numbers, something like, when we scroll to the section the animations goes like fast counter from 0 to destination number, what do you think @KonradSzwarc ?
> Here's simple example of what I was thinking about https://css-tricks.com/animating-number-counters/

Great idea 😄 I think you can create a separate task for this improvement, so we don't change the scope of this nearly finished task

KonradSzwarc (approved):

KrystianKjjk (approved):

PR #153, opened 2021-08-13
test(ui): create renderWithTheme function for tests (#123)

### LINKED ISSUE #123: Crete custom render function for tests
We want to create a custom `renderWithTheme` function that has the same interface as the default `render` from React Testing Library but wraps provided component in ThemeProvider

hoolek77:
Closes #123


packages/ui/src/utils/renderWithTheme.ts
```diff
@@ -0,0 +1,9 @@
+import { ReactElement } from 'react';
```
KonradSzwarc, line 1: 💭 For consistency with [another PR](https://github.com/CodersCrew/coderscamp/pull/126) add this function to `testHelpers.ts` file in the `src` directory

packages/ui/src/utils/renderWithTheme.ts
```diff
@@ -0,0 +1,9 @@
...
+// eslint-disable-next-line import/no-extraneous-dependencies
+import { render, RenderOptions, RenderResult } from '@testing-library/react';
+
+import { ThemeProvider } from '../theme';
+
+export const renderWithTheme = (ui: ReactElement, options: Omit<RenderOptions, 'queries'> = {}): RenderResult => {
```
KonradSzwarc, line 7: 🤔 This function should have the same interface the default `render` has. Here you omit the `queries` property and disallow to provide additional wrappers (they would overwrite the `ThemeProvider` if passed)

KonradSzwarc (changes requested):

packages/ui/src/testHelpers.tsx
```diff
@@ -0,0 +1,32 @@
...
+export const renderWithTheme = <
+  // eslint-disable-next-line @typescript-eslint/naming-convention
+  Q extends Queries = typeof queries,
+  Container extends Element | DocumentFragment = HTMLElement,
+>(
+  ui: React.ReactElement,
+  options: RenderOptions<Q, Container> = {},
+): RenderResult<Q, Container> => {
+  return render(<ThemeProvider>{ui}</ThemeProvider>, options);
+};
```
KonradSzwarc, line 32: 💭 As `render` is a function with type overload you can preserve it by writing code like the following:
```
const withTheme = <Render extends typeof render>(renderFc: Render) =>
  ((...args: Parameters<Render>) => {
    const [ui, options] = args;

    return renderFc(<ThemeProvider>{ui}</ThemeProvider>, options);
  }) as Render;

export const renderWithTheme = withTheme(render);
```
This way you also don't rely explicitly on current typings of separate render params and return types.

packages/ui/src/testHelpers.tsx
```diff
@@ -0,0 +1,32 @@
+/* eslint-disable no-console */
```
KonradSzwarc, line 1: 🤔 Why do we need this?

packages/ui/src/testHelpers.tsx
```diff
@@ -0,0 +1,32 @@
...
+
+import React from 'react';
+// eslint-disable-next-line import/no-extraneous-dependencies
```
KonradSzwarc, line 4: 💭 You can add `'**/testHelpers.tsx'` to `FILES_WITH_DEV_DEPENDENCIES` array in the `.eslintrc.js`

KonradSzwarc (changes requested):

packages/ui/src/testHelpers.tsx
hoolek77: This one comes from the other PR that implemented `withExpectedError`

KonradSzwarc (approved):

KonradSzwarc (approved):

KrystianKjjk (approved):

PR #159, opened 2021-08-14
feat(api): endpoint for initial form

kacper-cyra:


packages/api/prisma/schema.prisma
```diff
@@ -8,9 +8,31 @@ generator client {
...
+  email           String?
+  image           String
+  town            String?
+  gender          String?
+  birthYear       Int?
+  educationStatus String?
+  Survey          Survey?
+}
+
+model Survey {
```
KonradSzwarc, line 23: ❓ Shouldn't Survey have its own `id` property?

packages/api/prisma/schema.prisma
```diff
@@ -8,9 +8,31 @@ generator client {
...
 model User {
-  id       Int    @id @default(autoincrement())
-  githubId Int    @unique
-  fullName String
-  email    String
-  image    String
+  id              Int     @id @default(autoincrement())
+  githubId        Int     @unique
+  fullName        String?
+  email           String?
```
KonradSzwarc, line 14: 💭 I would consider making those properties empty strings if they are not set instead of making them nullable. It would allow us to have only one user type instead of specifying `User` and `RegisteredUser`

packages/api/prisma/schema.prisma
```diff
@@ -8,9 +8,31 @@ generator client {
...
+  image           String
+  town            String?
+  gender          String?
+  birthYear       Int?
+  educationStatus String?
```
KonradSzwarc, line 19: 💭 As I'm looking at the [task description](https://github.com/CodersCrew/coderscamp/issues/67#issue-947771156) only `sex` and `city` properties should be parts of the user model. After a short consideration, I think you can move even them to the `Survey` model

packages/api/src/surveys/survey.controller.spec.ts
```diff
@@ -0,0 +1,49 @@
...
+import { SurveyController } from './survey.controller';
+import { SurveyService } from './survey.service';
+
+describe('Survey controller', () => {
+  let controller: SurveyController;
+  let surveyRequest: SurveyPostRequest;
+  const userId = 1;
+
+  describe('saveUserSurvey', () => {
+    it('Returns void if request is Successful', async () => {
```
KonradSzwarc, line 16: 🙃 Successful could be lowercase

packages/api/src/users/users.service.ts
```diff
@@ -0,0 +1,21 @@
...
+
+@Injectable()
+export class UsersService {
+  constructor(private readonly usersRepository: UsersRepository) {}
+
+  async update(user: Partial<User> & UserFromJwt): Promise<User | RegisteredUser> {
+    const userFromDb = await this.usersRepository.getById(user.id);
+
+    if (!userFromDb) {
+      throw new Error('User not found');
```
KonradSzwarc, line 16: 💭 `throw new NotFoundError()` would be better than creating custom errors

packages/api/src/surveys/survey.service.ts
```diff
@@ -0,0 +1,26 @@
...
+    const { Survey: survey, ...user } = surveyData;
+    const surveyFromDb = await this.surveyRepository.findByUserId(user.id);
+
+    if (surveyFromDb) throw new BadRequestException(SurveyErrorMessage.SURVEY_ALREADY_FILLED);
+
+    const savedSurvey = await this.surveyRepository.save(survey);
+
+    const updatedUser = await this.usersService.update(user);
+
+    return [updatedUser, savedSurvey];
```
KonradSzwarc, line 24: 💭 It could be better to return an object from here so we can know immediately what those values stand for

packages/api/src/surveys/survey.mapper.ts
```diff
@@ -0,0 +1,41 @@
...
+    reasonForRetakingCourse,
+    expectations,
+    experience,
+    reasonToAccept,
+    plans,
+    absencePeriod,
+    averageTime,
+    associatedWords,
+    courseInformationSource,
+    marketingAccept,
```
KonradSzwarc, line 19: ❓ Do we need all the destructuring here?

packages/api/src/surveys/survey.controller.ts
```diff
@@ -0,0 +1,27 @@
...
+import { SurveyMapper } from './survey.mapper';
+import { SurveyService } from './survey.service';
+
+@Controller('surveys')
+export class SurveyController {
+  constructor(private surveyService: SurveyService) {}
+
+  @Post('survey')
+  async saveUserSurvey(
+    @Body() userSurveyRequest: SurveyPostRequest,
```
KonradSzwarc, line 17: 🤔 We should have some validator here to ensure data correctness

KonradSzwarc (changes requested):

packages/api/src/users/users.service.ts
nowakprojects: NotFound error is strictly HTTP exception, here we are on lower layer (business logic). NotFound is indeed better, but maybe in controller? Imagine that you're interact with this business logic by messaging or gRPC, then NotFound - HTTP 404 is not accurrate.

packages/shared/src/api/survey/surveyPostRequest.ts
```diff
@@ -0,0 +1,20 @@
...
+  email: string;
+  town: string;
+  birthYear: number;
+  gender: string;
+  educationStatus: string;
+  courseInformationSource: string[];
+  associatedWords: string[];
+  description: string;
+  prevParticipation: string;
+  reasonForRetakingCourse: string | null;
```
nowakprojects, line 12: Null or undefine? @KonradSzwarc what are best practices in TS? I'm in favor or undefined and avoid null whenever possible.

packages/shared/src/models/survey.ts
```diff
@@ -0,0 +1,19 @@
...
+  reasonToAccept: string;
+  plans: string;
+  absencePeriod: string;
+  averageTime: number;
+  associatedWords: string[];
+  courseInformationSource: string[];
+  marketingAccept: boolean;
+};
+
+export type CreateSurveyData = Omit<User, 'image'> & Survey;
```
nowakprojects, line 19: `Omit<User, 'image'>` do we want to change needed survey data whenever User changes? 
The fact that fields are duplicated, it's not the reason that change will always occur in those two places.

packages/api/src/auth/github/github.service.ts
```diff
@@ -9,7 +9,7 @@ import { UserFromGithub } from '../../users/users.types';
 export class GithubService {
   constructor(private readonly usersRepository: UsersRepository) {}
 
-  async authorizeUser(githubUserData: UserFromGithub): Promise<User> {
+  async authorizeUser(githubUserData: UserFromGithub): Promise<User | RegisteredUser> {
```
nowakprojects, line 12: I don't know why here is returned User OR RegisteredUser? All those property all important? Maybe just id or something like that can be returned?

packages/api/src/surveys/survey.controller.spec.ts
```diff
@@ -0,0 +1,49 @@
...
+      const response = httpMocks.createResponse();
+
+      await controller.saveUserSurvey(surveyRequest, userId, response);
+
+      expect(response._getStatusCode()).toBe(204);
+    });
+  });
+
+  async function setup() {
+    const module = await Test.createTestingModule({
```
nowakprojects, line 34: I recommend using include: [SurveyModule] and override providers which you want, because now you have totally different module definition that in prod (if you change prod code - tests will not be aware of that).

packages/api/src/users/users.service.ts
KonradSzwarc: The question is, wouldn't it be easier to care about other communication protocols when we actually plan to implement them?

nowakprojects: It depends... - as always :D 
But for me, it's similar like exposing database entity from rest api = mixing layers.

packages/api/src/surveys/survey.service.ts
nowakprojects: It always confuses me in default React Hooks. Do you know why they prefer tuples over object?

PR #163, opened 2021-08-16
feat(ui): create Sidebar and SidebarItem components (#10)

### LINKED ISSUE #10: Create Sidebar and SidebarItem components
**Acceptance criteria:**
- All states of SidebarItem and Sidebar match mockups.
- Sidebar item can receive `children`, `count` (optional).
- Can receive `icon` and `iconSelected` props (both or none).
- Avatar in the bottom of the Sidebar uses a re-exported Avatar component from Chakra UI.
- There are playground and states stories for the SidebarItem component.
- There is a story for the Sidebar component.

**Mockups:**
[SidebarItem](https://www.figma.com/file/ur0KLA9ZOtiodrAmpXrxh9/CodersCamp-app-and-website?node-id=9%3A13)
[Sidebar](https://www.figma.com/file/ur0KLA9ZOtiodrAmpXrxh9/CodersCamp-app-and-website?node-id=15%3A798)

kamildzierzak:
Closes #10


KrystianKjjk:
If "Navbar" component is website feature, sidebar shouldn't be on the website side too? 🤔

KonradSzwarc:
> If "Navbar" component is website feature, sidebar shouldn't be on the website side too? 🤔

Maybe not website but it should be on the panel side

packages/ui/src/components/SidebarItem/SidebarItem.test.tsx
```diff
@@ -0,0 +1,38 @@
...
+    render(
+      <SidebarItem path="/" disabled>
+        Dashboard
+      </SidebarItem>,
+      { wrapper: MemoryRouter },
+    );
+
+    const sidebarItem = screen.getByRole(/group/i);
+
+    expect(sidebarItem).toHaveStyle({ pointerEvents: 'none', opacity: '0.6' });
```
lagdush, line 36: At the moment we are testing if the element gets the right style. Maybe a better solution is to check if it gets the disabled attribute? As I checked, toBeDisabled () doesn't work here, toHaveAttribute ('disabled') works instead, but you have to block this line for eslint. @KonradSzwarc which solution is better?

hoolek77 (changes requested):
🙃  The sidebar item with the icon is missing some margin-left. Rest looks good for me 👏🏻.

hoolek77 (approved):
Looks good to me! ✅

lagdush (approved):

PR #169, opened 2021-08-17
feat(ui):creating EventsCard component in storybook with tests.

mrpodkalicki:


packages/ui/src/components/EventsCard/EventsCard.stories.tsx
```diff
@@ -0,0 +1,20 @@
+import React from 'react';
+import { Meta, Story } from '@storybook/react';
+
+import { EventCardProps, EventsCard } from './EventsCard';
+import { EVENTS as events } from './EventsCard.mocks';
+
+const meta: Meta = {
+  title: 'Components/EventsCard',
+  component: EventsCard,
```
KonradSzwarc, line 9: 💭 You can add `width` to the controls list and change its type to a number so it would be easier to customize it

packages/ui/src/components/EventsCard/EventsCard.mocks.ts
```diff
@@ -0,0 +1,33 @@
+export const MAIN_TITLE = 'Najbliższe terminy';
+
+export const BUTTON_TEXT = 'Przejdź do wydarzenia';
+
+const eventDescription =
+  'Sed dui elementum, elit est id malesuada leo varius. Arcu tempus, non justo leo egestas.' +
+  'A cras fames a ullamcorper. Laoreet ut orci semper massa integer ut consequat aliquam. Commodo enim, sed' +
+  'lorem nibh vivamus neque congue...';
+
+export const EVENTS = [
```
KonradSzwarc, line 10: 💭 As a convention we should use UPPER_CASE only for variables with primitive values

packages/ui/src/components/EventsCard/EventsCard.tsx
```diff
@@ -0,0 +1,91 @@
...
+
+export type Event = {
+  id: number;
+  title: string;
+  date: Date | { from: Date; to: Date };
+  description: string;
+  url: string;
+};
+
+export type EventCardProps = { events: Event[] } & BoxProps;
```
KonradSzwarc, line 17: 💭 As a convention we should use interfaces for component props

packages/ui/src/components/EventsCard/EventsCard.tsx
```diff
@@ -0,0 +1,91 @@
...
+    return '';
+  }
+
+  return `${addZeroToDateValue(date.getHours())}:${addZeroToDateValue(date.getMinutes())}`;
+};
+
+export const EventsCard = ({ events, ...props }: EventCardProps) => {
+  return (
+    <Box width="100%" borderRadius={8} padding="32px" boxShadow="base" backgroundColor="white" {...props}>
+      <Typography size="xl" color="gray.900" fontWeight={800} lineHeight="28px" mb="24px">
```
KonradSzwarc, line 46: 🤔 Typography shouldn't receive `fontWeight` (use `weight` instead) and `lineHeight` properties. I would need to look at why it was possible to pass them

packages/ui/src/components/EventsCard/EventsCard.tsx
```diff
@@ -0,0 +1,91 @@
...
+        {MAIN_TITLE}
+      </Typography>
+      {(events || []).map(({ id, title, date, description, url }: Event, index: number) => (
+        <Box
+          key={id}
+          display="flex"
+          flexDirection="row"
+          flexWrap="wrap"
+          w="100%"
+          mb={index !== events.length - 1 ? '32px' : '0px'}
```
KonradSzwarc, line 56: 💭 You can wrap events in a Stack to specify the spacing between them, so it wouldn't require a conditional

packages/ui/src/components/EventsCard/EventsCard.tsx
```diff
@@ -0,0 +1,91 @@
...
+  return `${addZeroToDateValue(date.getHours())}:${addZeroToDateValue(date.getMinutes())}`;
+};
+
+export const EventsCard = ({ events, ...props }: EventCardProps) => {
+  return (
+    <Box width="100%" borderRadius={8} padding="32px" boxShadow="base" backgroundColor="white" {...props}>
+      <Typography size="xl" color="gray.900" fontWeight={800} lineHeight="28px" mb="24px">
+        {MAIN_TITLE}
+      </Typography>
+      {(events || []).map(({ id, title, date, description, url }: Event, index: number) => (
```
KonradSzwarc, line 49: 🤔 When you want to pass a default value it would be better to do so when restructuring props

packages/ui/src/components/EventsCard/EventsCard.tsx
```diff
@@ -0,0 +1,91 @@
...
+        <Box
```
KonradSzwarc, line 50: 💭 For flexbox layout it would be better to use the Flex component

packages/ui/src/components/EventsCard/EventsCard.tsx
```diff
@@ -0,0 +1,91 @@
...
+  title: string;
+  date: Date | { from: Date; to: Date };
+  description: string;
+  url: string;
+};
+
+export type EventCardProps = { events: Event[] } & BoxProps;
+
+const addZeroToDateValue = (dateValue: number): string => {
+  if (dateValue === undefined) {
```
KonradSzwarc, line 20: ❓ If you made `dateValue` a required parameter why do you need to check it for `undefined`?

packages/ui/src/components/EventsCard/EventsCard.tsx
```diff
@@ -0,0 +1,91 @@
...
+  id: number;
+  title: string;
+  date: Date | { from: Date; to: Date };
+  description: string;
+  url: string;
+};
+
+export type EventCardProps = { events: Event[] } & BoxProps;
+
+const addZeroToDateValue = (dateValue: number): string => {
```
KonradSzwarc, line 19: 💭 You can probably use JavaScript's native `padStart` function instead of this one

packages/ui/src/components/EventsCard/EventsCard.tsx
```diff
@@ -0,0 +1,91 @@
...
+        >
+          <Typography data-testid="eventDate" size="5xl" lineHeight="48px" color="gray.900" fontWeight={400} mr="20px">
+            {date && 'from' in date ? getDayAndMonth(date.from) : getDayAndMonth(date)}
+          </Typography>
+          <Box maxW="570px">
+            <Typography data-testid="eventTitle" size="lg" lineHeight="28px" color="gray.900" fontWeight={700}>
+              {title}
+            </Typography>
+            <Typography data-testid="eventHours" size="sm" lineHeight="20px" color="gray.500" fontWeight={500} mb="8px">
+              {date && 'from' in date && 'to' in date
```
KonradSzwarc, line 67: 💭 Why do we check if `date` isn't falsy? Isn't it a required prop?

packages/ui/src/components/EventsCard/EventsCard.test.tsx
```diff
@@ -0,0 +1,247 @@
+import React from 'react';
+import { render, screen } from '@testing-library/react';
+
+import { Event, EventsCard } from './EventsCard';
+import { BUTTON_TEXT, EVENTS as listOfEvents, MAIN_TITLE } from './EventsCard.mocks';
+
+describe('EventsCard tests', () => {
```
KonradSzwarc, line 7: 🤔 It looks to me, that we're trying to test too many things here. In my opinion, you can create only one test to check if events render correctly and two tests to check date display correctness (for a single date and for the time period)

KonradSzwarc (changes requested):

packages/ui/src/components/EventsCard/EventsCard.tsx
mrpodkalicki: Hard to say, typescript show us a error ,,Type 'number' is not assignable to type 'undefined'' but storybook work well and text is bold.

packages/ui/src/components/EventsCard/EventsCard.tsx
mrpodkalicki: I think, even is this value is required we can't be sure that this value will not to be undefined.  Date.getDate() can return undefined so this condition is additional protection, but with your suggestion I used padStart so this function is no need.

packages/ui/src/components/EventsCard/EventsCard.tsx
mrpodkalicki: What do you mean by default value ? I don't use any default value for this props. Should I use ?

packages/ui/src/components/EventsCard/EventsCard.test.tsx
mrpodkalicki: Will be 7 tests, one for main title, one for properly render events, one for properly text in buttton, two for date (a single date and for the time period) and two for hours (single date and for the time period).

packages/ui/src/components/EventsCard/EventsCard.tsx
mrpodkalicki: it was needed for tests, when I tested undefined value, but this tests was remove so this check is no need also.

packages/ui/src/components/EventsCard/EventsCard.tsx
KonradSzwarc: When you destructure props, use `events = []`

packages/ui/src/components/EventsCard/EventsCard.tsx
mrpodkalicki: It's done

packages/ui/src/components/EventsCard/EventsCard.tsx
mrpodkalicki: it's done

packages/ui/src/components/EventsCard/EventsCard.tsx
mrpodkalicki: it's done

packages/ui/src/components/EventsCard/EventsCard.tsx
mrpodkalicki: it's done

packages/ui/src/components/EventsCard/EventsCard.mocks.ts
mrpodkalicki: it's done

packages/ui/src/components/EventsCard/EventsCard.stories.tsx
mrpodkalicki: it's done

packages/ui/src/components/EventsCard/EventsCard.mocks.ts
```diff
@@ -0,0 +1,33 @@
...
+  'Sed dui elementum, elit est id malesuada leo varius. Arcu tempus, non justo leo egestas.' +
+  'A cras fames a ullamcorper. Laoreet ut orci semper massa integer ut consequat aliquam. Commodo enim, sed' +
+  'lorem nibh vivamus neque congue...';
+
+export const events = [
+  {
+    id: 1,
+    title: 'Test kwalifikacyjny na CodersCamp',
+    date: { from: new Date(2021, 4, 4, 13, 40), to: new Date(2021, 5, 4, 15, 40) },
+    // we have to remember that month is number 0-11 not 1-12.
```
KonradSzwarc, line 15: 🙃 I think, we don't need to describe how JavaScript works

packages/ui/src/components/EventsCard/EventsCard.test.tsx
```diff
@@ -0,0 +1,129 @@
+import React from 'react';
+import { render, screen } from '@testing-library/react';
+
+import { EventsCard } from './EventsCard';
+import { BUTTON_TEXT, events as listOfEvents, MAIN_TITLE } from './EventsCard.mocks';
+
+describe('EventsCard tests', () => {
+  describe('render component tests', () => {
+    it('main title should be render', () => {
+      // when
```
KonradSzwarc, line 10: 🙃 Overall there is nothing bad in this practice, but we don't add such comments in other tests so you can remove yours for consistency

packages/ui/src/components/EventsCard/EventsCard.tsx
```diff
@@ -0,0 +1,75 @@
...
+}
+
+const getDayAndMonth = (date: Date): string => {
+  if (date === undefined) {
+    return '';
+  }
+
+  const month = date.getMonth() + 1;
+
+  return `${date.getDate()?.toString().padStart(2, '0')}.${month?.toString().padStart(2, '0')}`;
```
KonradSzwarc, line 29: ❓ Can `getDate` return `undefined`? Also can `month` be `undefined` when the `date` isn't?

packages/ui/src/components/EventsCard/EventsCard.tsx
```diff
@@ -0,0 +1,75 @@
...
+};
+
+const getHourAndMinute = (date: Date): string => {
+  if (date === undefined) {
+    return '';
+  }
+
+  return `${date.getHours()?.toString().padStart(2, '0')}:${date.getMinutes()?.toString().padStart(2, '0')}`;
```
KonradSzwarc, line 37: ❓ Can `getHours` and `getMinutes` return `undefined`?

packages/ui/src/components/EventsCard/EventsCard.test.tsx
```diff
@@ -0,0 +1,129 @@
...
+          description: '',
+          url: '#',
+        },
+      ];
+
+      // when
+      render(<EventsCard events={events} />);
+
+      // then
+      const eventDate = screen.getByTestId('eventHours');
```
KonradSzwarc, line 103: 💭 You can use `getByText` here, so there wouldn't be the need to use `data-testid`

packages/ui/src/components/EventsCard/EventsCard.test.tsx
```diff
@@ -0,0 +1,129 @@
...
+      });
+    });
+  });
+
+  describe('date tests', () => {
+    it('date should be render like 01.11 when date arg is new Date(2012, 10, 1, 10, 10)', () => {
+      // given
+      const events = [
+        {
+          id: 0,
```
KonradSzwarc, line 47: 💭 You can create a `createEventMock` function in the `EventsCard.mocks.ts` file. It will take the `overrides` param to customize particular mocks. This way you can reuse it in all places you create an event mock.

packages/ui/src/components/EventsCard/EventsCard.mocks.ts
```diff
@@ -0,0 +1,33 @@
+export const MAIN_TITLE = 'Najbliższe terminy';
+
+export const BUTTON_TEXT = 'Przejdź do wydarzenia';
```
KonradSzwarc, line 3: 🤔 Those constants aren't mocks and probably should be located in the `EventsCard.tsx` file

PR #170, opened 2021-08-17
feat(api): create checklist generation endpoint (#68)

### LINKED ISSUE #68: Create checklist generation endpoint
**Events flow:**
- Endpoint receives a request to generate a new Process.st checklist with the ID of the user requesting it.
- Server uses the provided ID to receive full name of the user.
- A bot is invoked.
- Bot goes to the link specified in `PROCESS_ST_URL` environment variable.
- Bot waits for the webpage to redirect to the address with a generated checklist.
- Bot returns the URL of the generated checklist.
- Returned URL is saved as `checklistUrl` property of the User entity.

KonradSzwarc:
Closes #68 

Co-authored-by: Mateusz Nowak <kontakt.mateusznowak@gmail.com>

### DIFF
--- a/packages/api/prisma/schema.prisma
+++ b/packages/api/prisma/schema.prisma
@@ -14,3 +14,10 @@ model User {
   email    String
   image    String
 }
+
+model LearningMaterial {
+  id             Int    @id @default(autoincrement())
+  url            String
+  completedCount Int    @default(0)
+  userId         Int    @unique
+}
--- a/packages/api/src/learning-materials/adapters/users.adapter.ts
+++ b/packages/api/src/learning-materials/adapters/users.adapter.ts
@@ -1,0 +1,19 @@
+import { Injectable, NotFoundException } from '@nestjs/common';
+
+import { UsersRepository } from '../../users/users.repository';
+import type { UsersPort } from '../ports/users.port';
+
+@Injectable()
+export class UsersAdapter implements UsersPort {
+  constructor(private readonly usersRepository: UsersRepository) {}
+
+  async getUserFullNameById(userId: number): Promise<string> {
+    const user = await this.usersRepository.getById(userId);
+
+    if (!user) {
+      throw new NotFoundException();
+    }
+
+    return user.fullName;
+  }
+}
--- a/packages/api/src/learning-materials/learning-materials.controller.ts
+++ b/packages/api/src/learning-materials/learning-materials.controller.ts
@@ -1,0 +1,26 @@
+import { Controller, Get, Post, UseGuards } from '@nestjs/common';
+
+import type {
+  CreateLearningMaterialResponse,
+  GetLearningMaterialResponse,
+} from '@coderscamp/shared/models/learning-material';
+
+import { JwtAuthGuard } from '../auth/jwt/jwt-auth.guard';
+import { UserId } from '../auth/jwt/user-id.decorator';
+import { LearningMaterialsService } from './learning-materials.service';
+
+@UseGuards(JwtAuthGuard)
+@Controller('learning-materials')
+export class LearningMaterialsController {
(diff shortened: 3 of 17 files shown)


packages/api/prisma/schema.prisma
```diff
@@ -14,3 +14,10 @@ model User {
   email    String
   image    String
 }
+
+model LearningMaterial {
+  id             Int    @id @default(autoincrement())
```
nowakprojects, line 19: Dajemy od razu string?

packages/api/src/learning-materials/learning-materials.controller.ts
```diff
@@ -0,0 +1,26 @@
...
+import { UserId } from '../auth/jwt/user-id.decorator';
+import { LearningMaterialsService } from './learning-materials.service';
+
+@UseGuards(JwtAuthGuard)
+@Controller('learning-materials')
+export class LearningMaterialsController {
+  constructor(private readonly learningMaterialsService: LearningMaterialsService) {}
+
+  @Get()
+  getLearningMaterial(@UserId() userId: number): Promise<GetLearningMaterialResponse> {
```
nowakprojects, line 18: Im not sure if it's LearningMaterial or Materials. Or maybe we should call it LearningMaterialsList?

nowakprojects (commented):
What about tests :) ? 
Testability is also a proof of good design.

packages/api/src/learning-materials/adapters/users.adapter.ts
```diff
@@ -0,0 +1,18 @@
+import { Injectable, NotFoundException } from '@nestjs/common';
+
+import { UsersRepository } from '../../users/users.repository';
+
+@Injectable()
+export class UsersAdapter {
```
nowakprojects, line 6: You can just call it Usernames, but UsersAdapter is also OK, but more technical.

nowakprojects: I haven't noticed that - in Ports & Adapters architecture adapters should implement interface, so rest of the code is dependent on abstraction and Adapter is a glue between what is needed (interface describe it) by module and how is provided.

nowakprojects: image

https://herbertograca.com/2017/11/16/explicit-architecture-01-ddd-hexagonal-onion-clean-cqrs-how-i-put-it-all-together/

packages/api/prisma/schema.prisma
KonradSzwarc: I think it would be better to convert all IDs to strings in a separate PR so an eventual discussion about such action could take place there. For feature PR I think it's better to just follow current conventions

packages/api/src/learning-materials/learning-materials.controller.ts
KonradSzwarc: I think LearningMaterial could be a good name as it allows us to potentially create more types of materials in the future. In such a case LearningMaterials would be reserved for an array o LearningMaterial elements

KonradSzwarc:
> What about tests :) ?
> Testability is also a proof of good design.

We plan to prepare our test environment to create tests on the controllers level (with real DB), so there is no need to create unit ones, where a lot of things would be mocked anyway.

nowakprojects:
> > What about tests :) ?
> > Testability is also a proof of good design.
> 
> @nowakprojects We plan to prepare our test environment to create tests on the controllers level (with real DB), so there is no need to create unit ones, where a lot of things would be mocked anyway.

I'm not into testing everything on controllers level - difficult to follow cohesion / single responsibility etc - it's easy to see with tests. 
But if it's a decision, so when :) ? It's similar like refactoring after merge. Probably will never happen :D

nowakprojects (approved):
For me lack of test is a crime, but beside of that everything looks OK :)

PR #171, opened 2021-08-18
refactor(api): separate UsersService and UsersRepository

KonradSzwarc:
A small change to increase the cohesion of our backend codebase

### DIFF
--- a/packages/api/src/auth/github/github.service.spec.ts
+++ b/packages/api/src/auth/github/github.service.spec.ts
@@ -3,7 +3,7 @@ import { Test } from '@nestjs/testing';
 import type { User } from '@coderscamp/shared/models/user';
 import { createObjectMock } from '@coderscamp/shared/utils/test';
 
-import { UsersRepository } from '../../users/users.repository';
+import { UsersService } from '../../users/users.service';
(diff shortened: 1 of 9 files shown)


kacper-cyra (approved):

packages/api/src/auth/github/github.service.spec.ts
```diff
@@ -37,28 +37,28 @@ describe('Github controller', () => {
...
 
-      usersRepository.getByGithubId = jest.fn().mockResolvedValue(dbUser);
+      usersService.getByGithubId = jest.fn().mockResolvedValue(dbUser);
 
       const githubUser = createObjectMock<UserFromGithub>({ githubId: dbUser.githubId });
 
       const result = await service.authorizeUser(githubUser);
 
-      expect(usersRepository.create).not.toHaveBeenCalled();
+      expect(usersService.create).not.toHaveBeenCalled();
```
nowakprojects, line 46: This is unnecessary. It's testing implementation details. It's not important in tests HOW something happened but WHAT. So `expect(result).toEqual(dbUser);` is enough.

nowakprojects (approved):

PR #172, opened 2021-08-18
refactor(api): convert all model IDs from number to uuid

KonradSzwarc:
We want to change the default ID type from sequence number to UUID as it would give us the following benefits:
-  we can generate id's anywhere (in other modules, on the frontend),
- we can use them in URLs without exposing the information about your data,
- it would make the app more future-proof (easier to scale or separate into smaller services).

### DIFF
--- a/packages/api/prisma/schema.prisma
+++ b/packages/api/prisma/schema.prisma
@@ -8,7 +8,7 @@ generator client {
 }
 
 model User {
-  id       String @id @default(uuid())
+  id       String @id @default(dbgenerated("gen_random_uuid()")) @db.Uuid
   githubId Int    @unique
   fullName String
   email    String
@@ -16,8 +16,8 @@ model User {
 }
 
 model LearningMaterial {
-  id             String @id @default(uuid())
+  id             String @id @default(dbgenerated("gen_random_uuid()")) @db.Uuid
   url            String
   completedCount Int    @default(0)
-  userId         String @unique
+  userId         String @unique @db.Uuid
 }
--- a/packages/api/prisma/schema.prisma
+++ b/packages/api/prisma/schema.prisma
@@ -8,7 +8,7 @@ generator client {
 }
 
 model User {
-  id       Int    @id @default(autoincrement())
+  id       String @id @default(uuid())
   githubId Int    @unique
   fullName String
   email    String
@@ -16,8 +16,8 @@ model User {
 }
 
 model LearningMaterial {
-  id             Int    @id @default(autoincrement())
+  id             String @id @default(uuid())
   url            String
   completedCount Int    @default(0)
-  userId         Int    @unique
+  userId         String @unique
 }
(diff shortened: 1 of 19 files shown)


kacper-cyra (approved):

packages/api/prisma/schema.prisma
```diff
@@ -8,16 +8,16 @@ generator client {
 }
 
 model User {
-  id       Int    @id @default(autoincrement())
+  id       String @id @default(uuid())
```
nowakprojects, line 11: I don't know Prisma. 
But do you think about @db.Uuid? Is it required / useful? 
Because as I see in migration.sql there is `TEXT`, whereas postgres have UUID type. I'd be great to use it for better indexing.

image

packages/api/src/auth/github/github.controller.spec.ts
```diff
@@ -4,13 +4,14 @@ import httpMocks from 'node-mocks-http';
...
 
+import { UserId } from '../../users/users.types';
 import { JwtService } from '../jwt/jwt.service';
 import { GithubController } from './github.controller';
 import { GithubService } from './github.service';
 import { GithubAuthGuardReq } from './github.types';
 
 const user = {
-  id: 1,
+  id: '1',
```
nowakprojects, line 14: Would be great to use in tests some real values, so here generate uuid. 
https://www.guidgenerator.com/

packages/api/src/auth/github/github.service.spec.ts
```diff
@@ -6,9 +6,10 @@ import { createObjectMock } from '@coderscamp/shared/utils/test';
 import { UsersService } from '../../users/users.service';
 import { UserFromGithub } from '../../users/users.types';
 import { GithubService } from './github.service';
+import type { GithubId } from './github.types';
 
-const createUser = (githubId: number): User => ({
-  id: Math.round(Math.random() * 100),
+const createUser = (githubId: GithubId): User => ({
+  id: String(Math.round(Math.random() * 100)),
```
nowakprojects, line 12: Those test may break, if random will return same value.

packages/api/src/auth/github/github.types.ts
```diff
@@ -2,8 +2,10 @@ import type { Request } from 'express';
 
 import type { GithubStrategy } from './github.strategy';
 
+export type GithubId = number;
+
 export interface GithubUser {
-  id: number;
+  id: GithubId;
```
nowakprojects, line 8: What do you think about `GithubUserId`? It's not id of Github, but user :D

packages/api/src/users/users.mapper.ts
```diff
@@ -14,13 +14,13 @@ export class UsersMapper {
 
   static fromJwtToDomain(payload: JwtPayload): UserFromJwt {
```
nowakprojects, line 15: I'd recommend to get rid of those mappers.

nowakprojects (approved):
Everything great, but force Prisma to change the type in Postgres from TEXT to UUID.

packages/api/src/users/users.mapper.ts
KonradSzwarc: Agree, but looks like an out-of-scope task from this PR perspective

PR #179, opened 2021-08-18
refactor: move to class-based validation (#178)

### LINKED ISSUE #178: Move to class-based validation
We plan to use class-validator and class-transformer to create shared validation models. In this task, we want to add those libraries to the project and use them to replace the current Joi-based environment variables validation.

KonradSzwarc:
Closes #178


packages/panel/vite.config.ts
```diff
@@ -2,10 +2,14 @@ import reactRefresh from '@vitejs/plugin-react-refresh';
...
+import { validateBuildEnv } from './src/common/validateBuildEnv';
 
-export default () => {
-  validateEnvVariables();
+export default async () => {
+  try {
+    await validateBuildEnv();
+  } catch (ex) {
+    throw new Error(ex);
+  }
```
KonradSzwarc, line 12: Those lines might look like they do nothing because we catch an error and throw it again. However, wrapping this exception in `Error` makes it possible to see the validation message. Below are two console outputs - the first is without the try-catch and the second with it.
CleanShot 2021-08-19 at 01 59 05@2x

nowakprojects (approved):

hoolek77 (approved):

PR #182, opened 2021-08-20
feat(api): move from GitHub auth to local auth (#181)

### LINKED ISSUE #181: Move from GitHub auth to local auth

KonradSzwarc:
Closed #181


packages/api/src/auth/auth.service.ts
```diff
@@ -0,0 +1,32 @@
...
+
+    if (current) {
+      throw new ConflictException(registerError.REGISTRATION_FORM_ALREADY_EXISTS);
+    }
+
+    const userRegistrationData = pick(data, ['fullName', 'email']);
+    const password = await hashPassword(data.password);
+
+    const { id } = await this.authRepository.createUserRegistration({ data: userRegistrationData });
+
```
kacper-cyra, line 27: What's the difference between userRegistration and authUser? Do we need them both as separate entities?

kacper-cyra (approved):

packages/api/src/auth/auth.service.ts
KonradSzwarc: You're right that `userRegistration` has no much use for now. I created it as we already know it will be needed when authentication flow expands by other steps (e.g. email confirmation #185). In such a case you will treat `userRegistration` as a bucket that stores data about the registration process until it finishes.

When it comes to `authUser` it's a table used by the entire auth module to perform auth-related operations. Doing it this way allows us to keep Auth module independent from any other modules.

packages/api/src/users/users.types.ts
```diff
@@ -1,6 +1,6 @@
 import { User } from '@coderscamp/shared/models/user';
 
-export type UserFromGithub = Pick<User, 'fullName' | 'githubId' | 'email' | 'image'>;
+export type UserFromGithub = Pick<User, 'fullName' | 'email' | 'image'>;
```
KrystianKjjk, line 3: shouldn't be 'UserFromGithub' renamed?  🤔

KonradSzwarc: No, it should be deleted 😄

packages/api/src/auth/auth.service.ts
nowakprojects: @KonradSzwarc - beautiful :)

packages/api/prisma/schema.prisma
```diff
@@ -9,10 +9,21 @@ generator client {
 
 model User {
```
nowakprojects, line 10: Is this name correct? Maybe UserPersonalInfo or something like that will be more accurate?

packages/api/src/auth/auth.controller.spec.ts
```diff
@@ -1,39 +1,91 @@
...
     controller = await module.resolve(AuthController);
   });
 
+  describe('register', () => {
+    it('calls `authService` with received body', async () => {
+      const bodyMock = { fullName: 'Some Name', email: 'some@mail.com', password: 'xyz' };
+
+      await controller.register(bodyMock);
+
+      expect(authService.register).toHaveBeenCalledWith(bodyMock);
```
nowakprojects, line 45: I'm not sure about this test. If service is mocked I'd rather define mocks behaviour in GIVEN, WHEN make a request, THEN assert with expect if controller response is conrrect based on GIVEN mocks.

packages/api/src/auth/auth.repository.ts
```diff
@@ -0,0 +1,16 @@
+import { Injectable } from '@nestjs/common';
+
+import { PrismaService } from '../prisma/prisma.service';
+
+@Injectable()
+export class AuthRepository {
+  constructor(private readonly prisma: PrismaService) {}
```
nowakprojects, line 7: I'd split repository by entity type = Single Responsibility

packages/api/src/auth/auth.service.ts
```diff
@@ -0,0 +1,32 @@
...
+import { hashPassword } from './local/local.utils';
+
+@Injectable()
+export class AuthService {
+  constructor(private readonly eventBus: EventBus, private readonly authRepository: AuthRepository) {}
+
+  async register(data: Prisma.UserRegistrationCreateInput & Prisma.AuthUserCreateInput) {
+    const current = await this.authRepository.findUserRegistration({ where: { email: data.email } });
+
+    if (current) {
```
nowakprojects, line 19: It may be true/false in line 19, but in 26, while we're creating user registration the state may be different. 
You can address it with database transaction isolation levels, or just do not check it in the code, but handle exception from database if email unique constraint may be violated.

packages/api/src/auth/auth.service.ts
```diff
@@ -0,0 +1,32 @@
...
+      throw new ConflictException(registerError.REGISTRATION_FORM_ALREADY_EXISTS);
+    }
+
+    const userRegistrationData = pick(data, ['fullName', 'email']);
+    const password = await hashPassword(data.password);
+
+    const { id } = await this.authRepository.createUserRegistration({ data: userRegistrationData });
+
+    await this.authRepository.createAuthUser({ data: { id, email: data.email, password } });
```
nowakprojects, line 28: I'd moved it out of that service (and split auth repository) and handle with event like UserRegistrationStartedEvent.

nowakprojects (changes requested):
Great work overall, but main logic is not tested (just controllers with mocked services).

packages/api/src/auth/auth.service.ts
```diff
@@ -0,0 +1,32 @@
...
+
+import { AuthRepository } from './auth.repository';
+import { UserRegistrationCompletedEvent } from './events/user-registration-completed.event';
+import { hashPassword } from './local/local.utils';
+
+@Injectable()
+export class AuthService {
+  constructor(private readonly eventBus: EventBus, private readonly authRepository: AuthRepository) {}
+
+  async register(data: Prisma.UserRegistrationCreateInput & Prisma.AuthUserCreateInput) {
```
nowakprojects, line 16: This is dangerous. Database logic spreads out all app layers.

nowakprojects: I'd pass to register some object like RegisterUserCommand and encapsulate all things related to db inside the service.

packages/api/src/auth/events/user-registration-completed.event.ts
```diff
@@ -0,0 +1,5 @@
+import type { UserRegistration } from '@prisma/client';
+
+export class UserRegistrationCompletedEvent {
+  constructor(public readonly payload: UserRegistration) {}
```
nowakprojects, line 4: Database entities should be internal to the module (you can change how you persist data whatever you want, but it should not impact other parts of an app). Here we publish database internals in events to whole application.

KonradSzwarc:
> Great work overall, but main logic is not tested (just controllers with mocked services).

@nowakprojects We spoke about it in another PR. Instead of making unit tests for all files separately, we plan to introduce e2e tests of the controllers

packages/api/src/auth/auth.controller.spec.ts
KonradSzwarc: I'm not sure if I understand correctly. Do you want to remove current expects and replace them with `expect(response).toBe(undefined);`?

packages/api/prisma/schema.prisma
KonradSzwarc: Great idea, but it's not in the scope of this PR

packages/api/src/auth/events/user-registration-completed.event.ts
KonradSzwarc: Isn't `UserRegistration` directly related to the module that sends this event? I mean this is an event that sends all information received by `UserRegistration`. In the current implementation if I for example add a `birthdate` field to registration form I wouldn't need to change it in many places

KrystianKjjk (approved):

packages/api/src/auth/events/user-registration-completed.event.ts
nowakprojects: You gave a specific example :) Adding filed is a good example of non-breaking operation. But assume that you remove field from database / split one into two, etc. What with modules which are subscribing for the event? The change will not be limited to just database and publisher, but to event and all subscribers. 
For me, good practice is to define module api (which are events part of - and in this architecture, service methods also) separately and treat database entities as implementation details. 
Another reason: i think it's really unexpected to change db access technology (like prisma to typeorm), but it sometimes happens, if yes - then this will have impact for almost all part of the app, so it's nice to do not leave module boundaries with internals like db access.

nowakprojects:
> > Great work overall, but main logic is not tested (just controllers with mocked services).
> 
> @nowakprojects We spoke about it in another PR. Instead of making unit tests for all files separately, we plan to introduce e2e tests of the controllers

For me code without tests is not completed. Testing is a must and a part of writing a feature. So example of such test and the infrastructure for that should be with the highest priority. You're not testing for your own risk. If it depends on me, I wouldn't sign myself under projects which is not covered by tests.

packages/api/src/auth/auth.controller.spec.ts
nowakprojects: Oh... hm... I haven't noticed that. But after what you said... I don't see value of this tests :P 
Maybe it shouldn't be discussed here, and leave testing for e2e.

PR #183, opened 2021-08-20
feat(website): create recruitment modals (#162)

### LINKED ISSUE #162: Create recruitment modals

lagdush:
closes #162


packages/website/src/components/Modals/ModalForm.tsx
```diff
@@ -0,0 +1,63 @@
...
+import { useBreakpointValue } from '@coderscamp/ui/hooks/useBreakpointValue';
+
+import { emailValidator } from '../Contact/Form';
+
+type ModalFormValues = {
+  name: string;
+  email: string;
+};
+
+const requiredValidator = {
```
KonradSzwarc, line 18: 💭 You can reuse the `requiredValidator` like you did with the `emailValidator`

packages/website/src/components/Modals/ModalForm.tsx
```diff
@@ -0,0 +1,63 @@
...
+      </FormControl>
+      <FormControl size={size} maxW={{ base: '100%', lg: '264px' }}>
+        <Input
+          {...register('email', { ...requiredValidator, ...emailValidator })}
+          invalid={Boolean(errors.email)}
+          placeholder="Adres e-mail"
+        />
+        {errors.email && <HelperText variant="error">{errors.email.message}</HelperText>}
+      </FormControl>
+      <Button type="submit" color="brand" variant="solid" size={size} weight="medium" p="10px 24px">
```
KonradSzwarc, line 58: 🤔 We shouldn't modify button properties like `weight` or `padding` as it made them incoherent with our design system

packages/website/src/components/Modals/ModalForm.tsx
```diff
@@ -0,0 +1,63 @@
...
+  } = useForm<ModalFormValues>();
+
+  const onSubmit: SubmitHandler<ModalFormValues> = (data) => {
+    // eslint-disable-next-line no-console
+    console.log(data);
+    reset();
+  };
+
+  return (
+    <Flex
```
KonradSzwarc, line 38: 💭 Maybe stack would be a little more suitable here as you want to achieve 16px of spacing between elements

packages/website/src/components/Modals/RecruitmentModal.tsx
```diff
@@ -0,0 +1,75 @@
...
+  modalText?: {
+    header: string;
+    body: string;
+    footer: string;
+  };
+};
+
+export const RecruitmentModal = ({ isOpen, onClose, modalText = forMentors }: RecruitmentModalProps) => {
+  return (
+    <>
```
KonradSzwarc, line 28: ❓ Why do we need this fragment?

packages/website/src/components/Modals/RecruitmentModal.tsx
```diff
@@ -0,0 +1,75 @@
...
+  isOpen: boolean;
+  onClose: () => void;
+  modalText?: {
+    header: string;
+    body: string;
+    footer: string;
+  };
+};
+
+export const RecruitmentModal = ({ isOpen, onClose, modalText = forMentors }: RecruitmentModalProps) => {
```
KonradSzwarc, line 26: 💭 It could be easier to understand if we always enforce to pass the `modalText` prop instead of making one of them the default

packages/website/src/components/Modals/RecruitmentModal.tsx
```diff
@@ -0,0 +1,75 @@
...
+          minHeight={448}
+          boxShadow="large"
+        >
+          <ModalHeader p={0}>
+            <Typography
+              as="h2"
+              size="4xl"
+              weight="extrabold"
+              color="gray.900"
+              textAlign="center"
```
KonradSzwarc, line 46: 💭 As all texts are aligned to the center, you can set this prop only once for the `ModalContent` element

packages/website/src/components/Modals/RecruitmentModal.tsx
```diff
@@ -0,0 +1,75 @@
...
+} from '@coderscamp/ui/components/Modal';
+import { Typography } from '@coderscamp/ui/components/Typography';
+
+import { forMentors } from './modal.data';
+import { ModalForm } from './ModalForm';
+
+type RecruitmentModalProps = {
+  isOpen: boolean;
+  onClose: () => void;
+  modalText?: {
```
KonradSzwarc, line 19: 💭 To be more DRY you can define an interface for modalText in modal.data.ts and only export it here.

packages/website/src/components/Modals/modal.data.ts
```diff
@@ -0,0 +1,12 @@
+export const forMentors = {
```
KonradSzwarc, line 1: 🙃 To follow our conventions for naming this file should be `Modal.data.ts` + the entire directory name should be `Modal` instead of `Modals`

packages/website/src/components/Schedule/Schedule.tsx
```diff
@@ -4,53 +4,61 @@ import { Button } from '@coderscamp/ui/components/Button';
 import { Center } from '@coderscamp/ui/components/Center';
 import { SimpleGrid } from '@coderscamp/ui/components/SimpleGrid';
 import { Typography } from '@coderscamp/ui/components/Typography';
+import { useDisclosure } from '@coderscamp/ui/hooks/useDisclosure';
```
KonradSzwarc, line 7: 🤔 To share our modals across sections we can add them to `_app.tsx` and then use context to display them from particular sections

KonradSzwarc (changes requested):

packages/website/src/components/Modals/ModalForm.tsx
lagdush: I did not export it because in issue the required text for the validator in the modal was "To pole nie może być puste". In the contact form we have "To pole jest wymagane". For consistency, I can export this validator and then we will have the same error information in the modal and the form.

KonradSzwarc: Great idea - re-export it so it would be easier for us to be consistent about this message

packages/website/src/components/Modals/RecruitmentModal.tsx
lagdush: I thought I would have to give these props to the modal but I solved it a little different now.

PR #184, opened 2021-08-20
feat(website): create benefits section (#97)

### LINKED ISSUE #97: Create Benefits section
**Mockups:**
[Figma link](https://www.figma.com/file/ur0KLA9ZOtiodrAmpXrxh9/CodersCamp-app-and-website?node-id=861%3A2893)

KrystianKjjk:
Closes #97


KonradSzwarc (changes requested):

packages/website/src/components/Benefits/AboutBenefits/AboutBenefitItems.data.tsx
```diff
@@ -0,0 +1,48 @@
...
+import {
+  BallotCheckIcon,
+  BooksIcon,
+  MessagesIcon,
+  SparklesIcon,
+  UserGroupCrownIcon,
+  UsersIcon,
+} from '@coderscamp/ui/icons';
+
+export const aboutBenefitItems: BenefitCardProps[] = [
```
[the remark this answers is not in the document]
KrystianKjjk: Hmm, I didn't noticed that all benefits are on figma. I will add all and implement expand button together

packages/website/src/components/Benefits/Benefits.tsx
```diff
@@ -0,0 +1,81 @@
...
+  text: 'Zwiń listę zalet CodersCamp',
+} as const;
+const rolledUpProfitButton = {
+  props: {
+    icon: <SolidArrowDownIcon />,
+  },
+  text: 'Rozwiń listę zalet CodersCamp',
+} as const;
+
+const BENEFITS_AFTER_SHRINKAGE = 6;
```
KonradSzwarc, line 32: 💭 It could be great to pass something like `shirinkSize` as a required prop instead of `showExpandProfitButton`. This way we can customize the number of displayed benefits per component usage. It would be useful when we for example decide that we want to sharing mentor benefits count but for a different number of tiles.

packages/website/src/components/Benefits/Benefits.tsx
```diff
@@ -0,0 +1,81 @@
...
+
+export const Benefits = ({ title, benefitItems, showExpandProfitButton = false }: BenefitsProps) => {
+  const columnsCount = useBreakpointValue({ base: 1, md: 2, xl: 3 } as const);
+  const profitButtonSize = useBreakpointValue({ base: 'sm', md: 'lg' } as const);
+  const [isExpanded, setIsExpanded] = useState(false);
+
+  const profitButtonValues = isExpanded ? expandedProfitButton : rolledUpProfitButton;
+  const handleProfitList = () => {
```
KonradSzwarc, line 40: 💭 `handleExpandButtonClick` could be a better name. Also, I think we should address the page jump after the user shrinks the list
CleanShot 2021-08-22 at 15 18 40

packages/website/src/components/Benefits/Benefits.tsx
KrystianKjjk: I changed that, with one exception, added shirinkSize as optional and every elements will be displayed if it won't be given

packages/website/src/components/Testimonials/Testimonials.data.ts
```diff
@@ -48,7 +48,7 @@ export const testimonials: Testimonial[] = [
   },
   {
     image: 'https://randomuser.me/api/portraits/women/88.jpg',
-    name: 'Darlene Mccoy',
+    name: 'Darlene Mccoy-wahad',
```
KonradSzwarc, line 51: 🙃 Is it a planned change?

KonradSzwarc (approved):

packages/website/src/components/Testimonials/Testimonials.data.ts
KrystianKjjk: xd 
When i copied data, I didn't change that name and this is the key prop... so, frustrating warning in console appeared

hoolek77 (approved):

PR #187, opened 2021-08-20
docs(api): add REST Client and create .http files for all controllers

KonradSzwarc:
In this PR I added `.http` files to describe API calls for all our controllers. They allow you to call all the endpoints directly from VS Code, when you install [this](https://marketplace.visualstudio.com/items?itemName=humao.rest-client) plugin and do "Rest Client: Switch Environment" command. Give it a try and write your opinion. Helper questions below:

1. Is it a good idea to add such files to our repository?
2. If yes, should we create one file per module, one in the API root, one/many but in another place?

P.S. This branch doesn't have `main` as the base branch for now as it's developed on top of the new auth.
P.P.S. If someone has any problems with running our backend, you can write to me on Slack.


nowakprojects:
Is it Visual Studio Code specific?

KonradSzwarc:
> Is it Visual Studio Code specific?

Yes

KrystianKjjk:
In my opinion it's a great idea to keep this file, it's definitely better than using postman. 

Maybe it will be better to keep one file per controller, but keep it in another folder, sth like 'apiCalls' directly in 'api'  like below?
image

and maybe keep in this .http file, name of the controller that the endpoints correspond to? 

In that case we will have every endpoints in one defined place, so it should be easier to find what our backend can do, and will be easier especially for people who will work only on frontend and will want to check one endpoint, won't be necessary to search .http file in the api tree

nowakprojects:
What about Swagger? Does it have any advantages over it?

KonradSzwarc:
> What about Swagger? Does it have any advantages over it?

I never used Swagger too much so here are my questions:
1. Does Swagger allow you to log in in case when you store JWT in cookies?
2. Does Swagger allow to define variables on the UI level (like @fullName in my files)?
3. Isn't it harder to set up and maintain Swagger than it is to write those files?

KrystianKjjk (approved):

PR #193, opened 2021-08-23
feat(website): create footer section (#101)

PiotrRynio:


packages/website/src/components/Footer/Footer.tsx
```diff
@@ -0,0 +1,100 @@
...
+      direction: useBreakpointValue({ base: 'column', lg: 'row' } as const),
+    } as const,
+    lowerFooter: {
+      direction: useBreakpointValue({ base: 'column', md: 'row' } as const),
+      order: useBreakpointValue({ base: -1, md: 1 } as const),
+    } as const,
+  } as const;
+
+  return (
+    <Flex w="100%" direction="column" align="center" bg={colors.gray['50']} color={colors.gray['500']} fontSize="16px">
```
KonradSzwarc, line 37: 💭 When you use colors you can write them simply as `"gray.50"`. Importing `colors` directly could be problematic as we cannot change those colors dynamically

packages/website/src/components/Footer/Footer.tsx
```diff
@@ -0,0 +1,100 @@
...
+];
+
+export const Footer = () => {
+  const footerStyles = {
+    upperFooter: {
+      direction: useBreakpointValue({ base: 'column', lg: 'row' } as const),
+    } as const,
+    lowerFooter: {
+      direction: useBreakpointValue({ base: 'column', md: 'row' } as const),
+      order: useBreakpointValue({ base: -1, md: 1 } as const),
```
KonradSzwarc, line 32: 💭 You don't have to keep those styles as objects. You can write them as:
```
const footerDirection = useBreakpointValue({ base: 'column', lg: 'row' } as const);
const lowerFooterOrder = useBreakpointValue({ base: 'column', lg: 'row' } as const);
```

packages/website/src/components/Footer/Footer.tsx
```diff
@@ -0,0 +1,100 @@
...
+
+            <Typography w="100%" pt={5} lineHeight="24px">
+              Organizatorem CodersCamp jest Stowarzyszenie CodersCrew - organizacja non-profit zrzeszająca pasjonatów
+              dziedzin związanych z IT.
+            </Typography>
+
+            <Flex pt={8}>
+              {socialMediaIcons.map((socialMediaIcon, index) => (
+                <Box key={socialMediaIcon.name}>
+                  {React.cloneElement(socialMediaIcon.component, { h: 6, w: 6, ml: index === 0 ? 0 : 7 })}
```
KonradSzwarc, line 51: 💭 Values provided with units would be better as there is no guessing on a reviewers side if there is a need to multiply them by 4 or not

packages/website/src/components/Footer/Footer.tsx
```diff
@@ -0,0 +1,100 @@
...
+
+        <Flex
+          justify="space-between"
+          borderTop="1px"
+          borderColor={colors.gray['200']}
+          minH="70px"
+          align="flex-end"
+          direction={footerStyles.lowerFooter.direction}
+        >
+          <Typography fontSizes="18px" lineHeight="28px" color={colors.gray['400']} pt="24px">
```
KonradSzwarc, line 90: 💭 For Typography you should use the `size` prop instead of providing `fontSize` or `lineHeight` manually. This way you will always ensure that your styles matches the design system

packages/website/src/components/Footer/Footer.tsx
PiotrRynio: I'm not sure If I understand what's different, but I'll make like you told

PiotrRynio: But why we have "colors.ts" file if we shouldn't use it?

packages/website/src/components/Footer/Footer.tsx
PiotrRynio: you are right! I'll do it!

packages/website/src/components/Footer/Footer.tsx
PiotrRynio: ok :))

packages/website/src/components/Footer/Footer.tsx
PiotrRynio: I think that object is easier for me and in future using.

packages/website/src/components/Footer/Footer.tsx
KonradSzwarc: When you write strings you use values from the theme object that is provided by context. In such a case when we change the theme (dynamically) all color tokens will be replaced by different values. It wouldn't be possible with a plain `colors` object.

colors.ts is a file when we define all colors accessible through the theme. You can track where they are imported and you will find our ThemeProvider

PR #194, opened 2021-08-23
refactor(api): replace ports-adapters with cqrs for learning-materials

KonradSzwarc:
I replaced the ports-adapters implementation used for learning materials creation flow with a query handled by the users' module. This way the functionality seems easier to grasp and we don't mix different concepts when it's not necessary.


packages/api/src/learning-materials/learning-materials.controller.ts
```diff
@@ -21,7 +32,19 @@ export class LearningMaterialsController {
   }
 
   @Post()
-  createLearningMaterial(@JwtUserId() userId: UserId): Promise<CreateLearningMaterialResponse> {
-    return this.learningMaterialsService.createLearningMaterial(userId);
+  async createLearningMaterial(@JwtUserId() userId: UserId): Promise<CreateLearningMaterialResponse> {
+    try {
```
nowakprojects, line 36: I suppose that such code will repeat in many places, so i'd be nice to disgn API like this: 

this.learningMaterialsService.createLearningMaterial(userId)
.onError(ResourceNotFoundException, e => throw new NotFoundException(learningMaterialError.USER_NOT_FOUND))
.onError(...)

What do you think about that? 
Maybe another task to standardize it somehow.

yarn.lock
```diff
@@ -2665,6 +2665,7 @@ __metadata:
   resolution: "@coderscamp/api@workspace:packages/api"
   dependencies:
     "@coderscamp/shared": 0.0.1
+    "@nestjs-architects/typed-cqrs": 0.1.1
```
nowakprojects, line 2668: What are the pros and cons of this library? 
As far as I see, it's not so popular.

packages/api/src/learning-materials/queries/what-is-user-name.query.ts
```diff
@@ -0,0 +1,9 @@
+import { Query } from '@nestjs-architects/typed-cqrs';
+
+import type { UserId } from '../../users/users.types';
+
+export class WhatIsUserNameQuery extends Query<string> {
```
nowakprojects, line 5: It should be in users module. 
It's dependency on this module, like subscribing for event from different module. 
I've just noticed is not exactly the same thing like ports & adapters, because there is no dependency inversion :D 
New questions (queries) for another modules will cause changes in module which have an answer :/ 
It twisted me before. Sorry :P Now I see Ports & Adapters may be better, 
but I think we can live with such solution too.

packages/api/src/learning-materials/queries/what-is-user-name.query.ts
```diff
@@ -0,0 +1,9 @@
+import { Query } from '@nestjs-architects/typed-cqrs';
+
+import type { UserId } from '../../users/users.types';
```
nowakprojects, line 3: Or.... maybe, what do you think - about placing query handler and this query in learning-materials... hmm....
Because the question is specific for learning materials, and query handler will work like an adapter.

packages/api/src/learning-materials/learning-materials.controller.ts
KonradSzwarc: I thought about some error handling standardization as well, so yes, let's do it in a separate task

yarn.lock
KonradSzwarc: Those are only types, so we don't need to worry. It gives us a possibility to define return types from queries and commands on a query/command level and then controls if you reaturn a correct type in a handler.

PR #195, opened 2021-08-24
feat(api): add role field to the AuthUser model (#166)

### LINKED ISSUE #166: Add `role` field to the AuthUser model
- AuthUser model should have a `role` field with an enum (ADMIN, USER) as a value.
- Each created user should have the USER role assigned by default.

KrystianKjjk:
Closes #166


packages/api/src/learning-materials/learning-materials.controller.ts
```diff
@@ -1,11 +1,11 @@
 import { Controller, Get, Post, UseGuards } from '@nestjs/common';
+import { JwtAuthGuard } from 'src/auth/jwt/jwt-auth.guard';
```
KonradSzwarc, line 2: 💭 Use a relative path as our tests and some scripts doesn't resolve paths that start with `src`

packages/api/src/guards/roles.Guard.ts
```diff
@@ -0,0 +1,28 @@
+import { CanActivate, ExecutionContext, Injectable, SetMetadata } from '@nestjs/common';
```
KonradSzwarc, line 1: 💭 Name of this file should be `roles.guard.ts` as we write all backend file names in kebab-case

KonradSzwarc (approved):

hoolek77 (approved):

PR #197, opened 2021-08-24
feat(website): add Real testimonials (#196)

### LINKED ISSUE #196: Add real testimonials

KonradSzwarc:
Closes #196


KonradSzwarc:
If someone considers why we needed to use some external service here. As testimonials contain personal data, everyone can ask us to remove it anytime. In such a case, it would be easier to remove them from the database than to run a rebase on the entire main branch.

Of course, if someone has a better solution for handling this situation, I'm open to ideas.

PR #233, opened 2021-08-27
fix: bigger buttons on the mobile site (#205)

### LINKED ISSUE #205: Bigger buttons on the mobile site
The "sign up", "become a mentor" and the menu button should be bigger on the mobile. The minimum required hover area of the element is 40x40px.

Screenshot_20210826_132035

lagdush:
Closes #205


wierzbka10 (commented):
Min hit area should be 40x40px, now menu icon have 32x32px and buttons have 32px high - size sm, shoud be md.

Veryyapeee (approved):

lagdush:
> Min hit area should be 40x40px, now menu icon have 32x32px and buttons have 32px high - size sm, shoud be md.
@wierzbka10 I understand this rule, but with the same button layout as we have, if the button size is larger, the entire page layout will be broken on the iPhone 5s, 6 and X. It will be ok for the iPhone 6 plus.

packages/website/src/components/Navbar/Navbar.tsx
```diff
@@ -24,11 +25,16 @@ export const Navbar = () => {
         position="relative"
       >
         {baseNavbar}
-        <Flex>
-          <Button size={buttonSize} mr="12px" onClick={() => openModal('mentor')}>
+        <Flex overflow="hidden">
+          <Button fontSize={isSmallerThan370px ? 'xs' : 'md'} size="md" mr="12px" onClick={() => openModal('mentor')}>
```
KonradSzwarc, line 29: 🤔 We shouldn't modify the font size of our buttons as it creates inconsistency in our design system

KonradSzwarc (changes requested):

wierzbka10:
@lagdush Could you show me in the video how it looks with larger buttons?

wierzbka10:
> 
> 
> > Min hit area should be 40x40px, now menu icon have 32x32px and buttons have 32px high - size sm, shoud be md.
> > @wierzbka10 I understand this rule, but with the same button layout as we have, if the button size is larger, the entire page layout will be broken on the iPhone 5s, 6 and X. It will be ok for the iPhone 6 plus.
> > I changed the font size on the smallest phones, check if it can stay that way.

@lagdush Could you show me in the video how it looks with larger buttons?

lagdush:
> > > Min hit area should be 40x40px, now menu icon have 32x32px and buttons have 32px high - size sm, shoud be md.
> > > @wierzbka10 I understand this rule, but with the same button layout as we have, if the button size is larger, the entire page layout will be broken on the iPhone 5s, 6 and X. It will be ok for the iPhone 6 plus.
> > > I changed the font size on the smallest phones, check if it can stay that way.
> 
> @lagdush Could you show me in the video how it looks with larger buttons?


https://user-images.githubusercontent.com/38866668/131379458-bd887124-3754-4028-aee4-b23fd92c2839.mov

packages/website/src/components/Navbar/Navbar.tsx
```diff
@@ -24,12 +25,12 @@ export const Navbar = () => {
         position="relative"
       >
         {baseNavbar}
-        <Flex>
-          <Button size={buttonSize} mr="12px" onClick={() => openModal('mentor')}>
+        <Flex overflow="hidden">
```
KonradSzwarc, line 28: ❓ Why do we need overflow to be hidden here?

lagdush: It seems to me that with overflow hidden, larger buttons do not destroy the page layout. This can be seen especially in the curriculum section of the iPhone 6 mobile view.

lagdush: But I can see that after deployment, the page looks different than for me locally.

wierzbka10 (changes requested):
I designed how shoud look buttons and hamburger button in the navigation on the different mobile sizes. On every screens smaller than 560px wide buttons "zapisz się na kurs" and "zostań mentorem" are inside drop-down menu, but on screens bigger or equel than 560px the buttons are in the navigation. Buttons in both are of the size md (40px high).

Link to figma: https://www.figma.com/file/ur0KLA9ZOtiodrAmpXrxh9/CodersCamp-app-and-website?node-id=2466%3A5013

wierzbka10 (changes requested):
buttons should be align to right, not centered.
buttons

wierzbka10 (approved):

hoolek77 (approved):

KonradSzwarc (approved):

PR #243, opened 2021-08-28
feat(website): add animation to website big numbers (#147)

### LINKED ISSUE #147: Add animation to website big numbers
- Add counter animation to big numbers in website

- Launch animation on scroll to section

Example: https://css-tricks.com/animating-number-counters/

kacper-cyra:
Closes #147


packages/ui/src/components/Counter/Counter.tsx
```diff
@@ -0,0 +1,62 @@
...
+function addThousandsSeparator(value: string) {
+  return value
+    .split('')
+    .reverse()
+    .join('')
+    .match(/.{1,3}/g)
+    ?.map((val) => val.split('').reverse().join(''))
+    .reverse()
+    .join(' ');
+}
```
lagdush, line 30: 💭 The animation looks great and I leave approve 👍  You can change the regular functions to the arrow functions that we basically use throughout the project.

lagdush (approved):

packages/ui/src/components/Counter/Counter.tsx
Veryyapeee: +1 for arrow functions

Veryyapeee (commented):
💭 I think that we could disable an animation on mobile devices, It's looking kinda weird for me, maybe one of the designers would refer to that

wierzbka10 (commented):
I feel a little confused looking at this animation, because some numbers move faster and other slowler so I don't know where to look and it makes a little difficult to read these informations. I'm not sure if this animation is needed here, but if you would love to do it, I would considered how we can lead user throught these numbers, so that he would pay attention to them and not feel overhelmed at the same time. For example You can try to swap the numbers so that they are from the smallest to the largest on one line and set the loading speed for each number to be the same. Then the numbers will stop sequentially and the user will follow the next numbers that will stop. I think that this can make user less overwhelmed.

wierzbka10:
> 
> 
> 💭 I think that we could disable an animation on mobile devices, It's looking kinda weird for me, maybe one of the designers would refer to that

I also think we could disable the animation on mobile devices because there the user sees the numbers separately and has to wait for them to load which slows him down and it can be a bit frustrating for him.

bartosznowosielski:
> 💭 I think that we could disable an animation on mobile devices, It's looking kinda weird for me, maybe one of the designers would refer to that

PR #246, opened 2021-08-28
fix: mobile menu is too small (#204)

### LINKED ISSUE #204: Mobile menu is too small
Expanded hamburger menu is not easily available on mobile screens.
It's too small - it shouldn't expand that way. Menu should be opened in fullscreen mode, it will be more accessible to the user and his little thumb.

lagdush:
Closes #204


packages/website/src/components/Navbar/BaseNavbar.tsx
```diff
@@ -65,3 +45,74 @@ export const DesktopBaseNavbar = () => {
     </>
   );
 };
+
+export const MobileBaseNavbar = () => {
+  const [isVisible, setIsVisible] = useState(false);
+  const logoLayout = useBreakpointValue({ sm: 'square', md: 'horizontal' } as const);
+  const { route } = useRouter();
+  const iconButonSize = 'md';
```
KonradSzwarc, line 53: 💭 I'm not sure if there is a need for this constant. In theory menu open and close buttons are separate components and doesn't have to share the same size

KonradSzwarc (changes requested):
There are a few problems with the current design:
1. Items don't look like they are clickable.
2. Items don't interact with the cursor. There is no pointer when you hover them, and they don't change their styles on hover.
3. Items are not focusable when you press Tab, or at least it's not visible that they are focusable.
4. The close button is too close to the left side.
5. If the current page is clickable (as it is now), it shouldn't have a distinguished styling. If the current page has a distinguished styling (as it has now), it shouldn't be clickable.

CleanShot 2021-08-29 at 23 01 02@2x

wierzbka10 (changes requested):
I designed menu on mobile :) On every screens smaller than 560px wide buttons "zapisz się na kurs" and "zostań mentorem" are inside drop-down menu, but on screens bigger or equel than 560px the buttons are in the navigation, not in the menu. Remember about hover and focus states. 

Link to figma: https://www.figma.com/file/ur0KLA9ZOtiodrAmpXrxh9/CodersCamp-app-and-website?node-id=2804%3A9699

wierzbka10 (changes requested):
You forgot about focus state on the clickable elements, now is only on buttons.

wierzbka10 (changes requested):
Focus state on clickable elements in menu shoud appear only when you move by tab like on the buttons

wierzbka10 (approved):

hoolek77 (approved):

KonradSzwarc (approved):

PR #254, opened 2021-08-31
feat(api): [read | LearningMaterials] enhance CourseProgress read model with handling TaskWasUncompleted (#229)

### LINKED ISSUE #229: [Read | LearningMaterials] Enhance CourseProgress read model with handling TaskWasUncompleted
**Dependencies (blockers)**
- [x] Make sure that #228 is done (must be merged before)
- [x] CourseProgress is already done

**Subtasks**
- [ ] Handle in `CourseProgressReadModule` event TaskWasUncompleted
- [ ] Update learningMaterialsCompletedTasks (decrease by one) CourseProgress entity on TaskWasUncompleted

KlaraGajaszek:
closes #229


packages/api/src/module/read/course-progress/course-progress.read-module.ts
```diff
@@ -25,4 +27,35 @@ export class CourseProgressReadModule {
       },
     });
   }
+
+  @OnEvent('LearningMaterialsTasks.TaskWasUncompleted')
+  async onTaskWasUncompleted(event: ApplicationEvent<TaskWasUncompleted>) {
+    const where = { learningMaterialsId: event.data.learningMaterialsId };
+    const courseProgress = await this.prismaService.courseProgress.findUnique({ where });
+
+    if (!courseProgress || courseProgress.learningMaterialsCompletedTasks === 0) {
```
nowakprojects, line 36: It's interesting. Very nice that you have spotted such edge case! :) 
We will think more carefully about such situations later (why something like that happened, if yes, what we should do? Such event probably should never come here? What about events ordering?), leave it as you have done for now. I'm working about handling logic right now, so maybe in the future we will change something :) 

Very good job!

packages/api/src/module/read/course-progress/course-progress.read-module.ts
```diff
@@ -25,4 +27,35 @@ export class CourseProgressReadModule {
...
+      return;
+    }
+
+    await this.prismaService.courseProgress.update({
```
nowakprojects, line 40: Whole method should be in transaction, but Prisma has better solution. Look how @KrystianKjjk has done it with increment. Do we have something silimar for decrement? Probably yes. 

In current implementation, when findUnique and update is asynchronous and not in the same transaction we may have situation that at line 34 we have completedTasks, let's say 5, but in line 43 it may be not true anymore (councurrent changes).

packages/api/src/module/read/course-progress/course-progress.read-module.ts
nowakprojects: I think this check is unnecessary, we can trust TaskWasUncompleted (is only for previously Completed task), because it's our internal event. 
And the check may also return false positives, because it's async - as I've stated in previous comment.

nowakprojects (approved):

packages/api/src/module/read/course-progress/course-progress.read-module.ts
nowakprojects: I think this check `if (!courseProgress || courseProgress.learningMaterialsCompletedTasks === 0)` is still unnecessary, we can trust TaskWasUncompleted (is only for previously Completed task), because it's our internal event.
And the check may also return false positives, because it's async - as I've stated in previous comment.

PR #256, opened 2021-08-31
feat(api): move CRUD based modules to crud directory

### LINKED ISSUE #255: Distinguish between CRUD modules and rich domain logic modules
Goal is to treat all modules as rich domain with unified architecture. 
But currently, some CRUD modules exists, so we want to have clear separation between them.

nowakprojects:
Closes #255 

I'm not sure where should be prisma module and guards.


packages/api/src/crud/courses/courses.controller.ts
```diff
@@ -2,8 +2,8 @@ import { Body, Controller, Get, Post, UseGuards } from '@nestjs/common';
 
 import { CourseBody, CreateCourseResponse, GetAllCoursesResponse } from '@coderscamp/shared/models/course';
 
-import { JwtAuthGuard } from '../auth/jwt/jwt-auth.guard';
 import { ForAdmin, RolesGuard } from '../auth/guards';
+import { JwtAuthGuard } from '../auth/jwt/jwt-auth.guard';
```
KonradSzwarc: 💭 We can use `@/auth` here

packages/api/src/module/read/learning-materials/learning-materials.read-module.spec.ts
```diff
@@ -6,9 +6,10 @@ import waitForExpect from 'wait-for-expect';
 import { initTestModule } from '@/common/test-utils';
 import { LearningMaterialsUrlWasGenerated } from '@/events/learning-materials-url-was-generated.domain-event';
 import { ApplicationEvent } from '@/module/application-command-events';
-import { UserId } from '@/users/users.types';
 import { EventStreamName } from '@/write/shared/application/event-stream-name.value-object';
 
+import { UserId } from '../../../crud/users/users.types';
```
KonradSzwarc, line 11: 💭 It would be great to add `crud` to TS paths

KrystianKjjk (approved):

packages/api/src/crud/courses/courses.controller.ts
nowakprojects: Not @/crud/auth if we want to have crud TS path?

PR #258, opened 2021-08-31
fix(website): react headroom - navigation (#202)

### LINKED ISSUE #202: React headroom - navigation
When you scroll down, navigation shoud slide out of view and slide back in when scrolling up like on this site https://kyleamathews.github.io/react-headroom/, beacuse it lets user in easy way come back to button with which he can sign up to the Coderscamp.

KrystianKjjk:
Closes #202


packages/website/src/components/Navbar/Navbar.tsx
```diff
@@ -8,13 +9,22 @@ import { useRecruitmentModal } from '@/components/RecruitmentModal';
 
 import { DesktopBaseNavbar, MobileBaseNavbar } from './BaseNavbar';
 
+const headroomStyles = {
+  zIndex: 1000,
+  background: 'rgba(255, 255, 255)',
+  boxShadow: '1px 1px 1px rgba(0,0,0,0.25)',
```
lagdush, line 15: ❓ Do we need box shadow here? I've noticed that in the mobile view, navigating behaves a bit strange.

https://user-images.githubusercontent.com/38866668/131527225-2ced7716-d51b-48ee-b843-cf2a0123b478.mov

packages/website/package.json
```diff
@@ -16,12 +16,14 @@
...
     "react": "17.0.2",
     "react-dom": "17.0.2",
+    "react-headroom": "3.1.1",
     "react-hook-form": "7.14.0",
     "swiper": "7.0.1"
   },
   "devDependencies": {
     "@types/nodemailer": "6.4.4",
     "@types/react": "17.0.19",
+    "@types/react-headroom": "^2.2.2",
```
KonradSzwarc, line 26: 🤔 We should pin all packages versions

packages/website/src/components/Navbar/Navbar.tsx
```diff
@@ -8,13 +9,22 @@ import { useRecruitmentModal } from '@/components/RecruitmentModal';
 
 import { DesktopBaseNavbar, MobileBaseNavbar } from './BaseNavbar';
 
+const headroomStyles = {
+  zIndex: 1000,
+  background: 'rgba(255, 255, 255)',
```
KonradSzwarc, line 14: 🤔 You should use values from our theme for `backgrond` and `boxShadow`. Also, box-shadow shouldn't appear when the navbar is in its initial position (at the top).

KonradSzwarc (changes requested):

packages/website/src/components/Navbar/Navbar.tsx
wierzbka10: I agree with Konrad and it looks better when is larger box-shadow, beacuse now it looks like simply line. In figma it's shadow style "large".

KrystianKjjk: @KonradSzwarc @wierzbka10  styling with colors like "gray.500" would not work, because its  not chakra component. I tried to connect chakra with headroom but it didn’t t work properly, so 'shadow="lg" ' does not work. 

What value of boxShadow should I use? I don't see it on figma nowhere 🤔

wierzbka10: @KrystianKjjk This shadow consists of two shadows:
shadow1
shadow2

lagdush (changes requested):
There is some bug with AboutTestimonials and Projects sections
https://user-images.githubusercontent.com/38866668/131892414-c393b86a-6e5e-4185-bf3c-f76e91481127.mov

KrystianKjjk:
> There is some bug with AboutTestimonials and Projects sections
> https://user-images.githubusercontent.com/38866668/131892414-c393b86a-6e5e-4185-bf3c-f76e91481127.mov

zIndex used in wrong place 
Now should be good :)

lagdush (approved):

wierzbka10 (approved):

KonradSzwarc (approved):

PR #295, opened 2021-09-04
feat(api): [write | emailConfirmation] create slice RequestEmailConfirmation -> EmailConfirmationWasRequested #279

### LINKED ISSUE #279: [Write | EmailConfirmation] create slice RequestEmailConfirmation -> EmailConfirmationWasRequested
**Context**
This implementation provides a generic approach to confirming any operation by email.
Current use case is user registration confirmation. But the same approach may be used in similar use cases like you approve bank account operations with sms code. 
Physical sending of email message by SMTP is not part of this task.

**Dependencies (blockers)**
- [ ] Make sure that #278 is done (must be merged before)

**Subtasks**
If you need more info or have questions, fell free to ask @nowakprojects.

Command definition:
- [ ] Create command type `RequestEmailConfirmation` in `modules/shared/commands`
- [ ] Create `RequestEmailConfirmationApplicationCommand` class in `modules/shared/commands`

Domain logic:
- [ ] Create function requestEmailConfirmation in `modules/write/email-confirmation/domain`.
- [ ] Return EmailConfirmationWasRequested from the method (no domain rules so far, will be introduced in next tasks).

Application logic:
- [ ] Create CommandHandler for RequestEmailConfirmationApplicationCommand
- [ ] Execute domain logic. Event stream name should be in format `EmailConfirmation_${userId}_${confirmationFor}` _Do you know why? Try to elaborate before taking this task._

Presentation logic:
- Rest API endpoint will be done in another task.

NestJS Module:
- [ ] Create EmailConfirmationWriteModule module and declare all required providers. Similar as other modules.

**Event Modeling**
MIRO link: https://miro.com/app/board/o9J_lQvnN28=/?moveToWidget=3074457363394933027&cot=14

KrystianKjjk:
Closes #279

### DIFF
--- a/packages/api/src/module/write/email-confirmation/domain/events.ts
+++ b/packages/api/src/module/write/email-confirmation/domain/events.ts
@@ -1,0 +1,3 @@
+import { EmailConfirmationWasRequested } from '@/module/events/email-confirmation-was-requested.domain-event';
+
+export type RequestEmailConfirmationDomainEvent = EmailConfirmationWasRequested;
--- a/packages/api/src/module/write/email-confirmation/domain/events.ts
+++ b/packages/api/src/module/write/email-confirmation/domain/events.ts
@@ -1,3 +1,3 @@
 import { EmailConfirmationWasRequested } from '@/module/events/email-confirmation-was-requested.domain-event';
 
-export type RequestEmailConfirmationDomainEvent = EmailConfirmationWasRequested;
+export type EmailConfirmationDomainEvent = EmailConfirmationWasRequested;
--- a/packages/api/src/module/write/email-confirmation/domain/requestEmailConfirmation.spec.ts
+++ b/packages/api/src/module/write/email-confirmation/domain/requestEmailConfirmation.spec.ts
@@ -1,0 +1,22 @@
+import { requestEmailConfirmationCommand } from '@/module/commands/request-email-conformation';
+
+import { requestEmailConfirmation } from './requestEmailConfirmation';
+
+describe('Request email confirmation', () => {
+  it('create event EmailConfirmationWasRequested when requestEmailConfirmation is invoked', () => {
+    const exampleUserId = 'ca63d023-4cbd-40ca-9f53-f19dbb19b0ab';
+    const exampleConfirmationToken = 'sbAPITNMsl2wW6j2cg1H2A';
+    const exampleConfirmationFor = 'user-registration';
+
+    const command = requestEmailConfirmationCommand({
+      userId: exampleUserId,
+      confirmationToken: exampleConfirmationToken,
+      confirmationFor: exampleConfirmationFor,
+    });
+    const result = requestEmailConfirmation(command)();
+
+    expect(result).toHaveLength(1);
+    expect(result[0].data).toEqual(command.data);
+    expect(result[0].type).toEqual('EmailConfirmationWasRequested');
+  });
+});
--- a/packages/api/src/module/write/email-confirmation/domain/requestEmailConfirmation.spec.ts
+++ b/packages/api/src/module/write/email-confirmation/domain/requestEmailConfirmation.spec.ts
@@ -1,9 +1,11 @@
 import { requestEmailConfirmationCommand } from '@/module/commands/request-email-conformation';
+import { emailConfirmationWasRequestedEvent } from '@/module/events/email-confirmation-was-requested.domain-event';
 
 import { requestEmailConfirmation } from './requestEmailConfirmation';
 
 describe('Request email confirmation', () => {
   it('create event EmailConfirmationWasRequested when requestEmailConfirmation is invoked', () => {
+    // given
     const userId = 'ca63d023-4cbd-40ca-9f53-f19dbb19b0ab';
     const confirmationToken = 'sbAPITNMsl2wW6j2cg1H2A';
     const confirmationFor = 'user-registration';
@@ -13,10 +15,13 @@ describe('Request email confirmation', () => {
       confirmationToken,
       confirmationFor,
     });
+
+    // when
     const result = requestEmailConfirmation(command)();
 
-    expect(result).toHaveLength(1);
-    expect(result[0].data).toEqual(command.data);
-    expect(result[0].type).toEqual('EmailConfirmationWasRequested');
+    // then
+    const event = emailConfirmationWasRequestedEvent(command.data);
+
+    expect(result).toEqual([event]);
   });
 });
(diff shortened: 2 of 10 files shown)


packages/api/src/module/write/email-confirmation/domain/events.ts
```diff
@@ -0,0 +1,3 @@
+import { EmailConfirmationWasRequested } from '@/module/events/email-confirmation-was-requested.domain-event';
+
+export type RequestEmailConfirmationDomainEvent = EmailConfirmationWasRequested;
```
nowakprojects, line 3: The purpose of this type union is to group together all events from one stream (related by some business rule - like `you cannot approve not requested confirmation` etc.), so better name would be `EmailConfirmationDomainEvent`. Currently there is only one `EmailConfirmationWasRequested`, but next `EmailConfirmationWasApproved` will be part of this union.

packages/api/src/module/write/email-confirmation/domain/requestEmailConfirmation.spec.ts
```diff
@@ -0,0 +1,22 @@
+import { requestEmailConfirmationCommand } from '@/module/commands/request-email-conformation';
+
+import { requestEmailConfirmation } from './requestEmailConfirmation';
+
+describe('Request email confirmation', () => {
+  it('create event EmailConfirmationWasRequested when requestEmailConfirmation is invoked', () => {
+    const exampleUserId = 'ca63d023-4cbd-40ca-9f53-f19dbb19b0ab';
```
nowakprojects, line 7: I think we don't need `example` before every value. 
Sometimes i follow approach like: 
```
const given = {
userId: ....
confirmationToken: ...
```

if you'd like, but just userId is enough. We know that we are in scope of test.

packages/api/src/module/write/email-confirmation/domain/requestEmailConfirmation.spec.ts
```diff
@@ -0,0 +1,22 @@
...
+    const exampleConfirmationFor = 'user-registration';
+
+    const command = requestEmailConfirmationCommand({
+      userId: exampleUserId,
+      confirmationToken: exampleConfirmationToken,
+      confirmationFor: exampleConfirmationFor,
+    });
+    const result = requestEmailConfirmation(command)();
+
+    expect(result).toHaveLength(1);
```
nowakprojects, line 18: It'd be better to check the whole result. Let's check out `given not generated before, then should be generated` test. 
Because now you haven't checked for example if there is no more event, or if event had properties that should not have (is TypeScript, so even if type doesn't allow that someone may be passed something more).

nowakprojects: AH, ok - you have `toHaveLength`, but I'm still in favor of checking equality. You can, because it's pure function (no side effects, output based only on input), so it's very easy to just check if result is as expected. It's more readable than splitting checks for different properties.

packages/api/src/module/write/email-confirmation/domain/requestEmailConfirmation.ts
```diff
@@ -0,0 +1,9 @@
+import { RequestEmailConfirmation } from '@/module/commands/request-email-conformation';
+import { emailConfirmationWasRequestedEvent } from '@/module/events/email-confirmation-was-requested.domain-event';
+
+import { RequestEmailConfirmationDomainEvent } from './events';
+
+export const requestEmailConfirmation =
```
nowakprojects, line 6: I'm not sure about this construct, that function return function. It is readable for you?

packages/api/src/module/write/email-confirmation/email-confirmation.spec.ts
```diff
@@ -0,0 +1,52 @@
...
+  it('creates EmailConfirmationWasRequested event when RequestEmailConfirmation has been executed', async () => {
+    // Given
+    const exampleUserId = 'ca63d023-4cbd-40ca-9f53-f19dbb19b0ab';
+    const exampleConfirmationToken = 'sbAPITNMsl2wW6j2cg1H2A';
+    const exampleConfirmationFor = 'user-registration';
+
+    // When
+    await moduleUnderTest.executeCommand(() => ({
+      class: RequestEmailConfirmationApplicationCommand,
+      ...requestEmailConfirmationCommand({
```
nowakprojects, line 35: Thanks for the proposition, it's a lot better than without such 'factory function' :)

packages/api/src/module/write/email-confirmation/email-confirmation.spec.ts
```diff
@@ -0,0 +1,52 @@
...
+    moduleUnderTest = await emailConfirmationTestModule();
+  });
+
+  afterEach(async () => {
+    await moduleUnderTest.close();
+  });
+
+  it('creates EmailConfirmationWasRequested event when RequestEmailConfirmation has been executed', async () => {
+    // Given
+    const exampleUserId = 'ca63d023-4cbd-40ca-9f53-f19dbb19b0ab';
```
nowakprojects, line 28: Same as before with the `example` part.

nowakprojects (changes requested):
I'm very proud of you! :D First task and done almost perfect!
Congratulations @KrystianKjjk.

packages/api/src/module/write/email-confirmation/domain/requestEmailConfirmation.spec.ts
nowakprojects: Yes, and my question is why don't just put in with one expect with whole expected output?

packages/api/src/module/write/email-confirmation/domain/requestEmailConfirmation.ts
nowakprojects: Ah, `requestEmailConfirmation(command)` will look nice :D Leave it as you have done.

packages/api/src/module/write/email-confirmation/domain/requestEmailConfirmation.spec.ts
KrystianKjjk: I don't know 🤷‍♀️ 
Now it is in one expect :)

nowakprojects (approved):

packages/api/src/module/write/email-confirmation/domain/requestEmailConfirmation.spec.ts
nowakprojects: For me it's easier to see whole object if there is test fail :) You have better context to understand problem.

KlaraGajaszek (approved):

PR #340, opened 2021-09-10
feat(api): create slice SendEmailMessage -> EmailMessageWasSent (#261)

### LINKED ISSUE #261: [Write | EmailSender] Create slice SendEmailMessage -> EmailMessageWasSent
**Dependencies (blockers)**
- No blockers. Ready to go!

**Subtasks**
Command definition:
- [x] Create command type `SendEmailMessage` in `modules/shared/commands`
- [x] Create `SendEmailMessageApplicationCommand` class in `modules/shared/commands`

Domain logic:
- [x] Create function sendEmailMessage in `modules/write/email-sender/domain`.
- [x] Take SendEmailMessage and appEmailAddress as parameters. appEmailAddress will be `from` of EmailMessageWasSent
- [x] Return EmailMessageWasSent on every SendEmailMessage.

Application logic:
- [x] Create env variable like APP_EMAIL_ADDRESS and use it as argument of sendEmailMessage.
- [x] Create CommandHandler for SendEmailMessageApplicationCommand
- [x] Send email using Nodemailer before saving EmailMessageWasSent event (executing domain logic). 
- [x] Execute domain logic for sendEmailMessage task. Event stream name should be in format `EmailMessage_${emailMessageId}`

Infrastructure
- [x] Configure node-mailer library. Try to use port & adapters architecture. Prepare in application layer port (interface) for sender and implement adapter which will invoke node-mailer. 

Testing strategy
- [x] Write tests for your module by invoking command and checking published event. 
- [ ] Test your command handler by executing a command and checking if mail was successfully sent. You can check if method of EmailSender was called or use Mailhog for that (add to docker-compose). Example test with mailhog you can find here: https://github.com/nowakprojects/CodersCamp2020.Project.FullStack-Node-React.TableSoccerTournaments/blob/develop/backend/test/modules/email-sending/core/application/SendEmail.mailhog-spec.ts

**Screenshot from MIRO (Event Modeling):**

CodersCamp App _ Event Storming - Frame 3 (3)

PawelSzambelan:
Closes #261


packages/api/src/module/write/email-sender/domain/sendEmailMessage.ts
```diff
@@ -0,0 +1,25 @@
+import {EmailMessageWasSentDomainEvent} from "@/write/email-sender/domain/events";
+import {SendEmailMessage} from "@/commands/send-email-message.domain-command";
+import {env} from "@/shared/env";
+
+export function sendEmailMessage(
+  _pastEvents: EmailMessageWasSentDomainEvent[],
+  command: SendEmailMessage,
+  appEmailAddress: env.APP_EMAIL_ADDRESS
```
PawelSzambelan, line 8: @all ktoś może zobaczyć, czemu mi tego nie łapię? Zrobiłem to wzorując się na innych importach z env, ale widocznie o czymś zapomniałem :-/ 
image
image

kacper-cyra: You are trying to use value as a type

nowakprojects: Tutaj daj jako string, bo typ jest string. A dopiero w wywołaniu przekaz ta wartość z env :)

nowakprojects:
Run `yarn w api lint:fix` to reformat your code :)

packages/api/src/module/write/email-sender/email-sending.spec.ts
```diff
@@ -0,0 +1,47 @@
...
+
+    // Then
+    await moduleUnderTest.expectEventsPublishedLastly<EmailMessageWasSentDomainEvent>([
+      {
+        type: 'EmailMessageWasSent',
+        data: { ...emailData, from: appEmail },
+        streamName: EventStreamName.from('EmailMessage',`EmailMessage_${emailMessageId}`)
+      }
+    ])
+
```
nowakprojects, line 45: Very good! Here add another assertion where you can test if emailSender.sendAnEmail was invoked with expected properties.

packages/api/src/module/write/email-sender/email-sending.write-module.ts
```diff
@@ -0,0 +1,10 @@
+import { Module } from '@nestjs/common';
+import {SharedModule} from "@/write/shared/shared.module";
+import {SendEmailMessageCommandHandler} from "@/write/email-sender/application/send-email-message.command-handler";
+
+@Module({
+  imports: [SharedModule],
+  controllers: [],
+  providers: [SendEmailMessageCommandHandler]
```
nowakprojects, line 8: There is lack of EmailSender provider. 
You can also override it for mocked implementation in tests.

packages/api/src/module/write/email-sender/infrastructure/node-mailer-email-sender.ts
```diff
@@ -0,0 +1,28 @@
+import {EmailSender} from "@/write/email-sender/application/email-sender";
+import Mail from "nodemailer/lib/mailer";
+import NodeMailer from 'nodemailer';
+
+export class NodeMailerEmailSender implements EmailSender {
+  private readonly transporter: Mail;
+  readonly mailFrom: string;
```
nowakprojects, line 7: I'm not sure if mailFrom should be here. More flexible solution may be to pass it as argument of sendAnEmail method (let's say we may want to send emails from mail like `rekrutacja@coderscamp.pl` / `mentorzy@coderscamp.pl` etc.) and use env variable in CommandHandler. 
What do you think? 
@KonradSzwarc - your opinion may be useful here.

packages/api/src/shared/env.ts
```diff
@@ -74,6 +74,31 @@ class EnvVariables {
   @Expose()
   @IsNumber()
   SUBSCRIPTION_QUEUE_WAITING_TIME_ON_RETRY_MS: number;
+
+  @Expose()
+  @IsString()
+  @IsNotEmpty()
+  APP_EMAIL_ADDRESS_TEST: string;
```
nowakprojects, line 81: What is it :) ?

nowakprojects (approved):

packages/api/src/module/write/email-sender/infrastructure/just-log.email-sender.ts
```diff
@@ -0,0 +1,14 @@
...
+import { EmailMessage, EmailSender } from '@/write/email-sender/application/email-sender';
+
+const logger = new Logger('JustLogEmailSender');
+
+export class JustLogEmailSender implements EmailSender {
+  constructor(private readonly next?: EmailSender) {}
+
+  async sendAnEmail(message: EmailMessage): Promise<void> {
+    logger.log('Email message was sent: ', message);
+    this.next?.sendAnEmail(message);
```
htk4, line 12: I think these two lines should be reordered or you can log before `sending` and after `sent`. You can also consider using fake but more realistic [email testing strategy](https://nodemailer.com/smtp/testing/) e.g. for presentation purposes.

htk4 (commented):
Overall LG :)

packages/api/src/module/write/email-sender/infrastructure/just-log.email-sender.ts
nowakprojects: You're correct - reordered :) 

Regarding better email testing strategy - I'm also into it. Something like that was the part of the task, but we will move it to another in order to do not stop dependent work. 
I'm not sure if Ethernal or MailTrap etc. is good for test on CI, because if there is something bad (outage for example) on the mail server, the tests of our application may break. So I think we can use some email client (like MailHog), which we can manage in Docker to have full controll over that.

Issue #337, opened 2021-09-09
Change buttons sizes on mobile


wierzbka10:
On the mobile every button shoud be md size. Below I show which butt
![image 1](https://user-images.githubusercontent.com/46423273/132760876-e8d98fc4-046f-4e6d-b304-13ca08f199df.jpg)
![image 2](https://user-images.githubusercontent.com/46423273/132760878-ee39688e-ffcf-46b7-bf05-ab13d620ca5b.jpg)
![image 3](https://user-images.githubusercontent.com/46423273/132760881-7e24cc1e-96fb-4aa0-8cc5-74220fa948eb.jpg)
![image 4](https://user-images.githubusercontent.com/46423273/132760885-a1aca3ec-15cb-4f2d-b63f-52d4f101f5b8.jpg)
ons shoud be change.

PR #344, opened 2021-09-12
fix(website): change buttons sizes on mobile (#337)

mrpodkalicki:


wierzbka10 (changes requested):
Buttons on the picture below 
butony mobilka długosc
should be the width of the box

PR #347, opened 2021-09-14
feat(api): open-api specification as contract testing tool (#225)

### LINKED ISSUE #225: [Presentation | REST API] Use OpenAPI specification as contract testing tool
**Context**
When we change something in the backend, we want to be sure that REST API still works as previous or any change in REST API contract is intended.
We won't generate swagger docs from TS code, but do it another way around.
So, we incorporate OPEN API specification from YAML file as API documentation and a tool for REST API contract testing.
It's also a perfect match for mocking with MSW for frontend testing:
https://dev.to/epilot/testing-react-with-jest-and-openapi-mocks-8gc

**Subtasks**
- [ ] Use supertest to make API Calls as described here: https://docs.nestjs.com/fundamentals/testing
- [ ] This test wouldn't be e2e. Mock commandBus in controllers.
- [ ] Test should check controller response on given CommandBus.execute result.
- [ ] Wrtie sample tests for rest api, which validate Open API contract. 

**Helpful resources**:
- Tests similar to that are expected (here project with Express.js):
https://github.com/nowakprojects/CodersCamp2020.Project.FullStack-Node-React.TableSoccerTournaments/blob/develop/backend/test/modules/doubles-tournament/presentation/rest-api/DoublesTournamentRestApiModule.spec.ts

nowakprojects:
Closes #225 

* Configure `openapi-jest` for contract testing.
* Example test for RestController.

### DIFF
--- a/packages/api/src/module/write/user-registration/presentation/rest/user-registration.rest-controller.spec.ts
+++ b/packages/api/src/module/write/user-registration/presentation/rest/user-registration.rest-controller.spec.ts
@@ -7,13 +7,7 @@ import { DomainRuleViolationException } from '@/shared/errors/domain-rule-violat
 import { initTestModuleRestApi } from '@/shared/test-utils';
 import { UserRegistrationRestController } from '@/write/user-registration/presentation/rest/user-registration.rest-controller';
 
-import { initOpenApiExpect } from '../../../../../../jest-setup';
-
-initOpenApiExpect();
-
-// fixme: ADDRESS ALREADY IN USE
-// eslint-disable-next-line jest/no-disabled-tests
-describe.skip('User Registration | REST API', () => {
+describe('User Registration | REST API', () => {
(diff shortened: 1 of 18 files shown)


packages/api/src/module/write/user-registration/presentation/rest/user-registration.rest-controller.spec.ts
```diff
@@ -0,0 +1,86 @@
...
+      });
+
+      // Then
+      expect(response.status).toBe(HttpStatus.BAD_REQUEST);
+      expect(response.body).toStrictEqual({
+        statusCode: HttpStatus.BAD_REQUEST,
+        message: 'USER_WAS_ALREADY_REGISTERED',
+        error: 'Bad Request',
+      });
+      expect(response).toSatisfyApiSpec();
```
nowakprojects, line 77: Does anyone know what to do for TS to recognize custom matchers? 
image

PR #373, opened 2021-09-24
fix(api): [EventsSubscription] retry for subcsription process - listen / catchup #370

### LINKED ISSUE #370: [EventsSubscription] Retry for subscription process - listen / catchup

htk4:
* Replace retry from ts-retry-promise lib with retryUntil - own implementation.

* Add graceful shutdown of EventsSubscription

Closes #370


packages/api/src/module/write/shared/application/events-subscription/events-subscription.spec.ts
```diff
@@ -38,7 +39,8 @@ describe('Events subscription', () => {
     subscription = sut.eventsSubscriptions
       .subscription(sut.randomUuid(), {
         start: { from: { globalPosition: 1 } },
-        retry: { retries: 3, backoff: 'LINEAR', delay: 50 },
+        retry: { backoff: 'FIXED', delay: 50, maxBackoff: 1000, resetBackoffAfter: 6 * 1000 },
```
nowakprojects, line 42: Do you have some better name to retry? Maybe something like `eventHandlingRetries`?

packages/api/src/module/write/shared/application/events-subscription/events-subscription.spec.ts
```diff
@@ -95,6 +97,7 @@ describe('Events subscription', () => {
       })
       .mockImplementationOnce(() => {}) // 7. event #4 - success (retry: 3)
       .mockImplementationOnce(() => {
+        subscription.stop(); // force stop processing - simulate no retries
```
nowakprojects, line 100: Why do we need it? Max retries for queue in config is 3. 
What is max for single event?

packages/api/src/module/write/shared/application/events-subscription/events-subscription.ts
```diff
@@ -89,43 +100,36 @@ export class EventsSubscription {
...
+
+      return;
+    }
+
+    this.logger.debug(`${this.subscriptionId} shutdown has been requested`);
     this.eventEmitter.offAny(this.eventEmitterListener);
-    this.queue.clear();
+    this.queue.stop();
+    this.running = false;
+    await this.shutdownToken.wait(60 * 1000);
```
nowakprojects, line 131: What is the purpose of that wait?

packages/api/src/module/write/shared/application/events-subscription/events-subscription.spec.ts
htk4: Too long :wink:

packages/api/src/module/write/shared/application/events-subscription/events-subscription.spec.ts
htk4: In reality, we have two retry strategies:

1. retry on queue in `listen()` phase to synchronize order of events ([short fixed delay](https://github.com/CodersCrew/coderscamp/pull/373/files#diff-549bcd44f6766e2dfd8c0ab2648924875a3972b0a52282e97796215f4120324eR231))
2. retry when sth goes wrong during events processing (e.g. DB connection lose, handleEvent throws an exception, etc.) ([configurable delay](https://github.com/CodersCrew/coderscamp/pull/373/files#diff-549bcd44f6766e2dfd8c0ab2648924875a3972b0a52282e97796215f4120324eR274-R295))

I removed number of retries from the second strategy. Now, EventsSubscriber works until specific condition. ( in our case: until stop/shutdown request ). I didn't wanna change this test so I decided to simulate stop processing after 3 retries.

packages/api/src/module/write/shared/application/events-subscription/events-subscription.ts
htk4: It's used to wait for current event processing to finish when the stop is requested. Without it, when the stop is requested, then the flow of the currently processing event can be unnecessarily broken. What's more, I think that some unresolved eventHandlers from one test can affect other tests. In that case ShutdownToken could improve stability of tests.

TL;DR
To gracefully shutdown background task ([this.run()](https://github.com/CodersCrew/coderscamp/pull/373/files#diff-549bcd44f6766e2dfd8c0ab2648924875a3972b0a52282e97796215f4120324eR114)).

nowakprojects (approved):
Nice job! :)

PR #375, opened 2021-09-25
fix(api): [EventsSubscription] should handle gaps in globalOrder #374

### LINKED ISSUE #374: [EventsSubscription] should handle gaps in globalOrder
## Problem description

The current implementation of `EventsSubscription` cannot work if there are gaps in `globalOrder`. Even though we can protect `globalOrder` sequence from gaps creation by using the `EventLock` table we can't secure the system from network partition. For example, consider a transaction that is running and has executed all statements but has not committed yet. Then if a network partition occurs between the database and the server, the transaction will abort, but the change made in `globalOrder` sequence will not rollback causing a gap.

htk4:
Closes #374


nowakprojects (changes requested):
I'm during review, but there are still some failing tests.

htk4:
> I'm during review, but there are still some failing tests.

This PR was waiting for that [PR](https://github.com/CodersCrew/coderscamp/pull/373). Now everything should work fine :smiley:

nowakprojects (approved):

